PR #6, opened 2020-02-13
LOOP-983 Updates for testing of first risk scenario

jh-bate:
covers card https://tidepool.atlassian.net/browse/LOOP-983


e2e/scenario_test/scenarios/Flat cgm.json
```diff
@@ -1,970 +1,1447 @@
 {
```
darinkrauss: [nit] I would do away with any spaces in filenames, if you can. (In recognize in some cases you cannot do so, such as in the directories associated with target names in Xcode.)

e2e/scenario_test/settingsTest.spec.js
```diff
@@ -8,32 +8,48 @@ describe('Pump Settings', () => {
...
+    });
     describe('Closed loop is not allowed', () => {
         describe('When correction range is not set', () => {
+            afterAll(async () => {
+                await cgm.removeData();
+                await pump.removeData();
+            });
             it('should not have correction range set', async () => {
-                await pump.checkCorrectionRange(false);
+                await pump.settings.CheckCorrectionRange(false);
```
darinkrauss: [nit] Consider pulling out `pump.settings` into just `settings`.  Otherwise, we'll end up with `pump.settings`, `cgm.settings`, and `settings` (for non-pump/cgm specific).

e2e/scenario_test/settingsTest.spec.js
```diff
@@ -8,32 +8,48 @@ describe('Pump Settings', () => {
...
-                await pump.setInsulinModel(pump.insulinModel.RapidAdults);
+                await pump.settings.InsulinModel(pump.insulinModel.RapidAdults);
             });
             it('should set the carb ratios', async () => {
-                await pump.setCarbRatios('8');
+                await pump.settings.CarbRatios('8');
             });
             it('should set insulin sensitivites set', async () => {
-                await pump.setInsulinSensitivities('500');
+                await pump.settings.InsulinSensitivities('500');
```
darinkrauss: [nit] I see some functions that start with lowercase and some with uppercase. Recommend on choosing one and using consistently, unless this difference means something.

e2e/scenario_test/settingsTest.spec.js
```diff
@@ -8,32 +8,48 @@ describe('Pump Settings', () => {
...
             });
-            it('should not allow closed loop mode', async () => {
+            it('should toggle on closed loop', async () => {
                 await setup.setClosedLoop();
```
darinkrauss: [nit] This is part of "settings", yes?

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
         RandomError: { value: 3, name: "Random Error" }
     },
     async add() {
-        await match.accessibilityButtonBarButton('Settings').tap();
-        if (element(by.text('Add CGM').and(by.type('UILabel')))) {
-            await element(by.text('Add CGM').and(by.type('UILabel'))).atIndex(0).tap();
-            await match.accessibilityButton('Simulator').tap();
-        }
-        await match.accessibilityButtonBarButton('Done').tap();
+        await match.accessible.ButtonBarButton('Settings').tap();
```
darinkrauss, line 32: I like how this is cleaning up, particularly the `match.accessible.X` stuff. It reads very nicely! 👍 

What is the difference between a "ButtonBarButton" and a "BackButton" (seen in `carbs.js`)?

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
...
+        await match.accessible.UILabel('Add CGM').tap();
+        await match.accessible.Button('Simulator').tap();
+        await match.accessible.ButtonBarButton('Done').tap();
     },
     async setEffect(effect) {
-        await match.accessibilityButtonBarButton('Settings').tap();
+        await match.accessible.ButtonBarButton('Settings').tap();
         //TODO assuming the second item as pump 'simulator' is first
```
darinkrauss, line 44: [nit] Address TODO?

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
...
-        await match.accessibilityLabelText('Simulator').atIndex(1).tap();
-        match.accessibilityHeaderText('CGM Settings');
-        await match.accessibilityLabelText(effect.name).tap();
+        await match.accessible.Label('Simulator').atIndex(1).tap();
+        match.accessible.HeaderText('CGM Settings');
```
darinkrauss, line 46: [nit] Do we want to `await` these checks?

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
...
     async add() {
-        await match.accessibilityButtonBarButton('Settings').tap();
-        if (element(by.text('Add CGM').and(by.type('UILabel')))) {
-            await element(by.text('Add CGM').and(by.type('UILabel'))).atIndex(0).tap();
-            await match.accessibilityButton('Simulator').tap();
-        }
-        await match.accessibilityButtonBarButton('Done').tap();
+        await match.accessible.ButtonBarButton('Settings').tap();
+        await match.accessible.UILabel('Add CGM').tap();
+        await match.accessible.Button('Simulator').tap();
```
darinkrauss, line 34: [nit] I wouldn't spend too much time/effort specifically testing either the pump or cgm simulator other than as a source of data for other tests. The simulators are only for our development/test use and will not ship with Tidepool Loop.

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
...
-                await element(by.type('LoopKitUI.PaddedTextField')).clearText();
-                await element(by.type('LoopKitUI.PaddedTextField')).typeText(bgValues[0]);
-                await match.accessibilityBackButton('Sine Curve').tap();
-                await match.accessibilityLabelText('Amplitude').tap();
-                await element(by.type('LoopKitUI.PaddedTextField')).clearText();
-                await element(by.type('LoopKitUI.PaddedTextField')).typeText(bgValues[1]);
-                await match.accessibilityBackButton('Sine Curve').tap();
-                await match.accessibilityBackButton('CGM Settings').tap();
+                await match.accessible.Label('Base Glucose').tap();
+                await match.uiEditableTextField().clearText();
```
darinkrauss: [nit] Is `uiEditableTextField` the currently active text field? If so, then might want to rename `uiEditableTextField` to something like `activeTextField` to make it clear that it is matching whatever field is active (and not one by other means) and so that the test writer needs to make a text field active before you can call `activeTextField`.

src/cgm.js
```diff
@@ -22,92 +22,85 @@ const cgm = {
...
-        //TODO static text and not a button?
-        await match.accessibilityLabelText('Delete CGM').tap();
-        await match.accessibilityLabelText('Delete CGM').atIndex(1).tap();
-        await waitFor(element(by.text('Add CGM').and(by.type('UILabel')))).toExist().withTimeout(2000);
-        await match.accessibilityButtonBarButton('Done').tap();
+        await match.accessible.ButtonBarButton('Settings').tap();
+        await match.accessible.Label('Simulator').tap();
+        match.accessible.HeaderText('CGM Settings');
+        await match.accessible.Label('Delete CGM').tap();
+        await match.accessible.Label('Delete CGM').atIndex(1).tap();
```
darinkrauss, line 105: [nit] Whenever I see `atIndex(n)` I wonder, does this mean there is *another* one someplace else? Since that is likely the case, consider adding a comment *why* you are using `atIndex` so it is clear to anyone who comes along later what is going on.

src/match.js
```diff
@@ -1,81 +1,90 @@
...
+         * @returns {Detox.Element} accessibilityButtonBarButton
+         */
+        ButtonBarButton(label) {
+            return accessibilityButtonBarButton(label, true);
+        },
+        /**
+         *
+         * @param {string} label
+         * @param {boolean} enabled
+         */
```
darinkrauss, line 19: [nit] `@returns`?

src/match.js
```diff
@@ -1,81 +1,90 @@
...
+        ButtonBarButton(label, enabled) {
+            accessibilityTraits = ['button'];
+            if (enabled == false) {
```
darinkrauss, line 22: [nit] If `enabled` is true, will this actually match *both* enabled and disabled? If so, consider renaming `enabled` to something like `onlyDisabled` to make it clear.

src/match.js
```diff
@@ -1,81 +1,90 @@
...
-     * @returns {Detox.Element} accessibilityBackButton
-     */
-    accessibilityBackButton(label) {
-        return element(by.label(label).and(by.traits(['button']).and(by.type('UIAccessibilityBackButtonElement'))));
-    },
     /**
      * @param {string} label
      * @returns {Detox.Element} buttonBarButton
      */
     buttonBarButton(label) {
```
darinkrauss: [nit] Will this be going away, too?

src/match.js
```diff
@@ -1,81 +1,90 @@
...
+            return element(by.text(text).and(by.traits(['text'])));
+        },
     },
     /**
      * @summary get a text field element
      * @returns {Detox.Element}
      */
-    textField() {
+    uiEditableTextField() {
         return element(by.type('LoopKitUI.PaddedTextField'));
```
darinkrauss, line 82: [nit] Not sure if all text fields in Loop will conform to this type.

src/match.js
```diff
@@ -1,81 +1,90 @@
...
-            accessibilityTraits.push('disabled');
-        }
-        return element(by.label(label).and(by.traits(accessibilityTraits).and(by.type('_UIButtonBarButton'))));
+    accessible: {
+        /**
+         * @param {string} label
+         * @returns {Detox.Element} accessibilityButtonBarButton
+         */
+        ButtonBarButton(label) {
+            return accessibilityButtonBarButton(label, true);
```
darinkrauss: [nit] Where is this defined?

src/pump.js
```diff
@@ -2,6 +2,190 @@ const element = require('detox').element;
...
 
 const pump = {
+    settings: {
+        /**
+         * @name pump.settings.BasalRates
+         * @param {string} [basalUnitsPerHour] e.g. '0.1'
+         */
+        async BasalRates(basalUnitsPerHour) {
+            const unitsSuffix = 'U/hr';
+            await match.accessible.ButtonBarButton('Settings').tap();
```
darinkrauss: [nit] Is there any way to make tapping the Settings button and the Done button (at the end) part of a "before/after each"? Not sure how that comes into play here (if it even does or if that is only for tests).

src/setup.js
```diff
@@ -24,26 +24,41 @@ const setup = {
...
+            await expect(match.accessible.Button('Closed Loop')).toHaveValue('1');
         }
-        await match.accessibilityButtonBarButton('Done').tap();
+        await match.accessible.ButtonBarButton('Done').tap();
+    },
+    /**
+     * @name unsetClosedLoop
+     * @summary turn off closed loop mode
+     */
+    async unsetClosedLoop() {
```
darinkrauss: [nit] `openLoop`?

darinkrauss (approved):
A number of nits and TODOs, but up to you whether you address them. Since nothing critical, LGTM. 🚀

e2e/scenario_test/settingsTest.spec.js
jh-bate: yeah agreed - I have done this now

e2e/scenario_test/scenarios/Flat cgm.json
jh-bate: done - they where generated that way but I agree

e2e/scenario_test/settingsTest.spec.js
jh-bate: all consistent now

e2e/scenario_test/settingsTest.spec.js
jh-bate: yes - i have moved now

src/match.js
jh-bate: yeah that was a mistake, fixed now

src/pump.js
jh-bate: possible - will think about that as I continue to develop things

src/setup.js
jh-bate: yeah thats a better name ;)

PR #8, opened 2020-02-26
Sprint11 setup

jh-bate:
@darinkrauss I am just going to move in these settings changes first so keen just to get this unit of work reviewed and then I will merge


e2e/scenario_test/settings/not_set/basal_rates/settings.no_basal_rates.spec.js
```diff
@@ -1,47 +1,24 @@
-const { setup, pump, cgm, match, settings } = require('../../../../../src/index');
+const { setup, match, settings, loopSettings } = require('../../../../../src/index');
```
darinkrauss, line 1: [nit] Perhaps one thing that could help the "../../.." issue would be to get rid of the deep directory hierarchy and focus on a standard naming scheme for the tests. So, perhaps this file would be "e2e/scenario_test/settings.not_set.basal_rates.spec.js". Also, consider that each of these test files may be common to one or more "groups" of tests (scenario vs. smoke).

e2e/scenario_test/settings/valid/settings.valid.spec.js
```diff
@@ -1,48 +1,23 @@
...
+                scenario: 'flat_cgm',
+                settings: settings.Filter(settings.Defaults)
+            };
+            await loopSettings.Configure(config);
+        });
+        it('should not be in closed loop mode yet', async () => {
+            await expect(match.loop.Icon()).toHaveLabel('Waiting for first run');
+        });
+        it.skip('should show no alert when tapping loop icon', async () => {
+            //TODO: need a valid way to assert we are in closed loop mode
```
darinkrauss, line 19: [nit] TODO

src/loopSettings.js
```diff
@@ -0,0 +1,22 @@
...
+     * @summary Configures loop with simulator cgm, simulator pump and all required settings applied to give a running instance of loop
+     * @param {object} config  { scenario string, settings object }
+     * @example await settings.ConfiguredLoop({scenario:'flat_cgm', settings: settings.Defaults})
+     */
+    async Configure(config) {
+        await setup.LoadDeviceScenariosFromDisk(device.id);
+        await cgm.AddSimulator();
+        await pump.AddSimulator();
+        await setup.LoadScenario(config.scenario);
+        await settings.Apply(config.settings);
```
darinkrauss, line 15: [nit] Not sure if this makes sense, but I would think that the scenario would be loaded after the settings were set. (In fact, I'd have to look at the scenario code, but it might be required.)

src/loopSettings.js
```diff
@@ -0,0 +1,22 @@
...
+    },
+};
+
+
+module.exports = loopSettings;
```
darinkrauss, line 20: [nit] No newline at end of file.

src/loopSettings.js
```diff
@@ -0,0 +1,22 @@
...
+const loopSettings = {
+    /**
+     * @summary Configures loop with simulator cgm, simulator pump and all required settings applied to give a running instance of loop
+     * @param {object} config  { scenario string, settings object }
+     * @example await settings.ConfiguredLoop({scenario:'flat_cgm', settings: settings.Defaults})
+     */
+    async Configure(config) {
+        await setup.LoadDeviceScenariosFromDisk(device.id);
+        await cgm.AddSimulator();
+        await pump.AddSimulator();
```
darinkrauss: [nit] Would it make sense to have the simulators as part of the settings.Apply? For any of these tests, will the simulators require a different configuration (e.g. CGM simulator requires a different BG generation choice)?

src/settings.js
```diff
@@ -19,7 +49,9 @@ const settings = {
      */
     async BasalRates(rates) {
         const unitsSuffix = 'U/hr';
-        await match.accessible.ButtonBarButton('Settings').tap();
+        if (startedFromHomeScreen){
```
darinkrauss: [nit] Space between `){`.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+         * @summary  PreMealCorrectionRange: { min: '179', max: '180' }
+         */
+        PreMealCorrectionRange: { min: '179', max: '180' },
+        /**
+         *  @summary  ClosedLoop: true
+         */
+        ClosedLoop: true,
+    },
+    Type:{
+        BasalRates:'BasalRates',
```
darinkrauss: [nit] Missing space.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+    /**
+     * @summary helper function to set seetings by applying the {settings.Defaults}
+     * @param  {Defaults} values list of settings that will not be applied
+     * @example await settings.Apply(settings.Defaults)
+     */
+    async Apply(values) {
+        startedFromHomeScreen = false;
+        await goToSettingsScreen();
+        if (values.BasalRates) {
+            await this.BasalRates(values.BasalRates);
```
darinkrauss: [nit] Consider pushing the check for not-nil into the function itself.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+        }
+        if (values.DeliveryLimits) {
+            await this.DeliveryLimits(values.DeliveryLimits);
+        }
+        if (values.SuspendThreshold) {
+            await this.SuspendThreshold(values.SuspendThreshold.threshold);
```
darinkrauss: [nit] Why the extra `.threshold` at the end?

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+        if (values.CarbRatios) {
+            await this.CarbRatios(values.CarbRatios);
+        }
+        if (values.InsulinSensitivities) {
+            await this.InsulinSensitivities(values.InsulinSensitivities);
+        }
+        if (values.CorrectionRanges) {
+            await this.CorrectionRanges(values.CorrectionRanges);
+        }
+        if (values.ClosedLoop) {
```
darinkrauss, line 434: [nit] Consider making this a boolean that will do both open and close the loop.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+        }
+        return filtered;
+    },
+    /**
+     * @summary helper function to set seetings by applying the {settings.Defaults}
+     * @param  {Defaults} values list of settings that will not be applied
+     * @example await settings.Apply(settings.Defaults)
+     */
+    async Apply(values) {
+        startedFromHomeScreen = false;
```
darinkrauss, line 425: [nit] The `startedFromHomeScreen` seems a bit "hackish". Consider an alternate solution to "remember" what view you are on. Perhaps a stack that you push and pop views from. Not sure exactly how it would work, but this way seems fragile. Just a thought.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+        /**
+         *  @summary  ClosedLoop: true
+         */
+        ClosedLoop: true,
+    },
+    Type:{
+        BasalRates:'BasalRates',
+        CarbRatios:'CarbRatios',
+        DeliveryLimits:'DeliveryLimits',
+        SelectInsulinModel:'SelectInsulinModel',
```
darinkrauss: [nit] Consider removing `Select` since we are just setting the `InsulinModel`.

src/settings.js
```diff
@@ -227,7 +290,111 @@ const settings = {
...
+         */
+        SelectInsulinModel: { value: 2, name: 'Rapid-Acting – Children' },
+        /**
+         * @summary CarbRatios: [{ time: '12:00 AM', carbGramsPerInsulinUnit: '8' }]
+         */
+        CarbRatios: [{ time: '12:00 AM', carbGramsPerInsulinUnit: '8' }],
+        /**
+         * @summary InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }]
+         */
+        InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }],
```
darinkrauss, line 370: [nit] Consider making the defaults "reasonable" values. Many of these are already, but `InsulinSensitivities`, `CorrectionRanges`, and `PreMealCorrectionRange` seem unlikely.

darinkrauss (approved):
A number of nits for the next go around, but otherwise LGTM. 🚀

src/settings.js
jh-bate: I agree - most of this is still a WIP as it comes down to actually being able to select the more reasonable values. This will happen when we have accessible pickers

src/settings.js
jh-bate: yeah I agree ... i am still mulling over the best way to do it. This works for now but will need changing in the near future

PR #10, opened 2020-03-10
create status class for accessing status screen items

jh-bate:
- expand smoke test to use status screen items
- use status screen functions to simplify risk based tests
- formatting cleanup

@darinkrauss this should cleanup a few of those nit pics


e2e/regression_test/settings.no_basal_rates.spec.js
```diff
@@ -1,8 +1,10 @@
-const { setup, match, loopSettings, FilterSettings, SettingDefault, SettingType } = require('../../src/index');
+const { setup, loopSettings, Status, FilterSettings, SettingDefault, SettingType } = require('../../src/index');
 
 describe('Closed loop is not allowed when settings', () => {
+    var status;
     beforeAll(async () => {
         await setup.LaunchLoop();
+        status = new Status();
```
darinkrauss: [nit] Perhaps `loopStatus` to mirror `loopSettings`?  Otherwise `status` is not too specific.

e2e/smoke_test/smokeTest.spec.js
```diff
@@ -1,29 +1,41 @@
...
         carbs = new Carbs();
         bolus = new Bolus();
         overrides = new Overrides();
+        status = new Status();
     });
-    describe('main screen', () => {
+    describe('status screen', () => {
         it('has Active Carbohydrates section', async () => {
-            await expect(match.accessible.Label('Active Carbohydrates')).toExist();
+            await status.OpenActiveCarbohydratesChart();
```
darinkrauss: [nit] I'm confused, why would `status` open the active carbohydrates chart? Perhaps this is a naming thing?

darinkrauss (approved):
LGTM. Just a couple of naming issues perhaps, but looks good! ⭐️

e2e/regression_test/settings.no_basal_rates.spec.js
jh-bate: hmmm this is for functions relating to the "status" screen. Should i rename all to be `StatusScreen` `SettingsScreen` etc ?

darinkrauss: Is the status screen the home screen? The top part of the home screen? Not sure what?

jh-bate: I thought so, in that case I can just rename it as `HomeScreen`

PR #12, opened 2020-03-20
LOOP-1026 Jhbate/sprint12

jh-bate:


e2e/accessibility_test/accessibility.spec.js
```diff
@@ -0,0 +1,190 @@
...
+            await expect(screen.carbEntry.DateLabel()).toExist();
+        });
+        it('has a Amount Consumed Label', async () => {
+            await expect(screen.carbEntry.AmountConsumedLabel()).toExist();
+        });
+        it('has a Food Type Label', async () => {
+            await expect(screen.carbEntry.FoodTypeLabel()).toExist();
+        });
+    });
+    describe.skip('override', () => {
```
darinkrauss, line 94: [nit] I do not believe overrides will be part of Tidepool Loop 1.0.

e2e/accessibility_test/accessibility.spec.js
```diff
@@ -0,0 +1,190 @@
...
+            await expect(screen.settings.PumpHeader()).toExist();
+        });
+        it('has a pump header', async () => {
+            await expect(screen.settings.ContinuousGlucoseMonitorHeader()).toExist();
+        });
+        it('has a settings header', async () => {
+            await expect(screen.settings.SettingsHeader()).toExist();
+        });
+        it('has a Issue Report Label', async () => {
+            //TODO: this is acting like a button!
```
darinkrauss, line 140: [nit] Just pointing out a bunch of TODOs here.

e2e/accessibility_test/accessibility.spec.js
```diff
@@ -0,0 +1,190 @@
+const { setup, screen } = require('../../src/index');
+
+describe('accessibility', () => {
+    beforeAll(async () => {
+        await setup.LaunchLoop();
+    });
+    describe('home', () => {
+        it('has a Active Carbohydrates Label', async () => {
```
darinkrauss, line 8: [nit] Is there anyway to shorten this syntax? Not sure. It is repeated a lot.
```
async () => {
  await expect(<something>).toExist();
}
```
This might be the shortest possible form given Javascript, but thought I'd ask if there was anything shorter that still made sense (and wasn't cryptic).

e2e/accessibility_test/accessibility.spec.js
```diff
@@ -0,0 +1,190 @@
...
+        it('has a Add Pump label', async () => {
+            //TODO: this is acting like a button!
+            await expect(screen.settings.AddPumpLabel()).toExist();
+        });
+        it('has a Add CGM label', async () => {
+            //TODO: this is acting like a button!
+            await expect(screen.settings.AddCGMLabel()).toExist();
+        });
+        it('has a services header', async () => {
+            await screen.settings.ScrollToBottom();
```
darinkrauss, line 186: [nit] How will device size for testing affect scrolling? Will we need more scrolling in other places on smaller devices?

e2e/regression_test/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,32 @@
+const { setup, loopSettings, screen, SettingDefault, CGMModel } = require('../../src/index');
+
+describe.skip('Closed loop is stopped when we have cgm data loss', () => {
+    let config = {
+        scenario: 'flat_cgm_trace',
+        settings: SettingDefault,
+    };
+    beforeAll(async () => {
+        await setup.LaunchLoop();
```
darinkrauss, line 9: [nit] Is there any way to push the `beforeAll` and `afterAll` globally? I imagine we need to run these two for *any* test. (Or is that not the case?)

src/bolusScreen.js
```diff
@@ -8,20 +8,47 @@ class BolusScreen {
     async Open() {
         try {
             //assume we are starting from the open screen
-            await match.accessible.Button(BolusLabel.Bolus).tap();
+            await match.accessible.Button(text.bolusScreen.Bolus).tap();
```
darinkrauss, line 11: [nit] For all of these `Open` and `Close` functions for the various screens, is there a way we can ensure that it actually opened or closed properly and ended up in the final expected location? I see here we are waiting until the bolus button is tapped, but we haven't actually confirmed that we opened the bolus screen itself. Or is that confirmation part of the tests that call the `Open` and `Close`? To me it would make sense to confirm here so we know if `Open` or `Close` returned successfully then we can be certain we are in the expected state. Thoughts?

src/carbEntryScreen.js
```diff
@@ -8,21 +8,94 @@ class CarbEntryScreen {
     async Open() {
         try {
             //assume we are starting from the open screen
-            await match.accessible.Button(CarbsLabel.AddMeal).tap();
+            await match.accessible.Button(text.carbEntryScreen.AddMeal).tap();
```
darinkrauss, line 11: [nit] I see that you've separated the "screen" implementation from the text that is expected to be on the screen. Consider putting anything related to a particular screen (whether it be actions such as `Open` or `Close`, controls such as `CancelHeaderButton`, or text) within the screen class. That way everything for a particular screen is in one place.

In other words, consider moving the contents of `text.carbEntryScreen` into this class, perhaps at the bottom. 

OTOH, I see you are also using "general" text labels, too, like `text.general.Cancel`.

Well, something for you to consider. :)

src/carbEntryScreen.js
```diff
@@ -8,21 +8,94 @@ class CarbEntryScreen {
...
+    DateLabel() {
+        return match.accessible.Label(text.carbEntryScreen.Date);
+    }
+    FoodTypeLabel() {
+        return match.accessible.Label(text.carbEntryScreen.FoodType);
+    }
+    AbsorptionTimeLabel() {
+        return match.accessible.Label(text.carbEntryScreen.AbsorptionTime);
+    }
+    /**
```
darinkrauss, line 38: [nit] I would drop most of these comments like this. I don't see them adding much useful, and, at least in this case, it is wrong. ;)

darinkrauss: I'd only add comments if it adds information that isn't already provided by the name of the function. Perhaps if there is a caveat or special considerations. That sort of thing.

src/carbEntryScreen.js
```diff
@@ -8,21 +8,94 @@ class CarbEntryScreen {
...
-        await match.accessible.ButtonBarButton(Label.Continue).tap();
-        await match.accessible.Button(CarbsLabel.SaveWithoutBolusing).tap();
+        await this.SaveWithoutBolusButton().tap();
+    }
+    /**
+     * @param {string} glucoseValueAndUnits
+     * @example await carbs.ExpectPredictedGlucoseWarning('110 mg/dL');
+     */
+    async ExpectPredictedGlucoseWarning(glucoseValueAndUnits) {
+        const predictedGlucoseWarning = `⚠ Predicted glucose of ${glucoseValueAndUnits} is below your suspend threshold setting.`;
```
darinkrauss, line 85: [nit] I like your idea of putting all of the "text" in one location (at least per class). Consider moving this text into that.

src/carbEntryScreen.js
darinkrauss: OTTH (On The Third Hand), if your intention of putting all of the text strings into one file (or one location, at least) so that you can easily swap out localizations, then I fully support that. 👍 Might need to change the implementation of `text.***` to something that can dynamically load the correct localization somehow, but that could be done at a later date.

src/homeScreen.js
```diff
@@ -1,33 +1,90 @@
 const match = require('./match');
-const { Label, HomeLabel } = require('./labels');
+const text = require('./text');
 
 class HomeScreen {
+    /**
+     * @example home.ActiveCarbohydratesLabel().tap();
+     */
+    ActiveCarbohydratesLabel() {
```
darinkrauss, line 8: 💯 I like your best practice of the suffix indicating what kind of thing it is (e.g. `Label`, `Button`).

src/loopSettings.js
```diff
@@ -1,27 +1,51 @@
 const setup = require('./setup');
-const { SettingsScreen } = require('./settingsScreen');
+const { SettingsScreen, SettingType, FilterSettings } = require('./settingsScreen');
 
 
+function filterSettingsBasedOnConfig(config) {
+    if (config.scenario) {
+        //CGM data will be loaded from scenario and we have already added the CGM simulator so remove that
+        return FilterSettings(config.settings, [SettingType.CGMSimulatorSettings, SettingType.AddCGMSimulator, SettingType.AddPumpSimulator]);
```
darinkrauss, line 8: Good catch!

src/settingsScreen.js
```diff
@@ -117,56 +122,172 @@ var FilterSettings = function (values, types) {
...
 };
 
+
+var _setCGMEffect = async function (effect) {
+    await match.accessible.Label(effect).tap();
+    switch (effect) {
+        case CGMEffect.GlucoseNoise:
+            await match.UIEditableTextField().clearText();
+            await match.UIEditableTextField().typeText('100');
+            //TODO:
```
darinkrauss: TODO what?

src/settingsScreen.js
```diff
@@ -117,56 +122,172 @@ var FilterSettings = function (values, types) {
     return filtered;
 };
 
+
+var _setCGMEffect = async function (effect) {
+    await match.accessible.Label(effect).tap();
+    switch (effect) {
+        case CGMEffect.GlucoseNoise:
+            await match.UIEditableTextField().clearText();
+            await match.UIEditableTextField().typeText('100');
```
darinkrauss, line 131: [nit] Consider making these constants into parameters?

src/settingsScreen.js
```diff
@@ -357,165 +478,134 @@ class SettingsScreen {
...
+            await match.accessible.Label(text.general.Save).tap();
+            await _exitSetting();
         }
-
     }
     /**
-     * @name settings.ClosedLoop
      * @summary turn on closed loop mode
      */
     async ClosedLoop() {
```
darinkrauss: [nit] Since this is taking an action to turn on closed loop mode, might want to consider making the function name action oriented (e.g. `SetClosedLoop` or something like that). If a function is mutating state (changing something) then I find it easier to understand if the function name is an action (e.g. Set, Do, Load, Close, Open, etc.).

darinkrauss (approved):
Sweet, looking good. A number of nits that are minor things for you to consider going forward. Otherwise it is really coming together! 🥇

e2e/accessibility_test/accessibility.spec.js
jh-bate: yeah - these will become jira cards to update the loop app

e2e/accessibility_test/accessibility.spec.js
jh-bate: yeah true - I hadn't thought of that

e2e/accessibility_test/accessibility.spec.js
jh-bate: I haven't fully investigated yet - but I am using a header within the app and then going right to the bottom. Issue will be if that header is off the screen. I plan on testing against a known list of devices

src/carbEntryScreen.js
jh-bate: agreed

e2e/accessibility_test/accessibility.spec.js
jh-bate: 👍 that makes sense

src/homeScreen.js
jh-bate: cheers

src/carbEntryScreen.js
jh-bate: yeah it was for localisation purposes.

src/bolusScreen.js
jh-bate: i could open and then assert I can see the screens header

PR #13, opened 2020-03-30
LOOP-1063 LOOP-1062 LOOP-1061 LOOP-1060 Jhbate/sprint13

jh-bate:
- guardrails tests 
- common 'loop' root for all app interaction 
- simplify setup 
- update detox and jest versions


e2e/regression_guardrail/guardrail.correction_range_max.spec.js
```diff
@@ -0,0 +1,16 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail correction range maximum is set to 180', () => {
```
darinkrauss: [nit] What should this be testing? That you can set it to 180 or that cannot set it to 181 or *both*?

e2e/regression_guardrail/guardrail.correction_range_max_warning.spec.js
```diff
@@ -0,0 +1,16 @@
+const { loop } = require('../../src/index');
+
+describe('guard rail correction range maximum warning is set to 120', () => {
```
darinkrauss: [nit] I'm not sure what this is testing. The filename implies that there should be a warning, but I don't see that in the code. Plus, testing boundary conditions is good (e.g. 180 vs. 181 and 60 vs 59), but not sure what testing a middle value like 120 does. What am I missing?

e2e/regression_guardrail/guardrail.correction_range_min.spec.js
```diff
@@ -0,0 +1,16 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail set correction range minimum is set to 60', () => {
```
darinkrauss: [nit] Consider testing both 60 works and 59 does not.

e2e/regression_guardrail/guardrail.correction_range_max_warning.spec.js
darinkrauss: Oh, I think I remember that above 120 and below 70 should pop a warning. Is that right? If so, then, I guess we'd have 8 tests, right?

- 59 - cannot be set
- 60 - can be set, warning
- 69 - can be set, warning
- 70 - can be set, no warning
- 120 - can be set, no warning
- 121 - can be set, warning
- 180 - can be set, warning
- 181 - cannot be set

Is that right?

darinkrauss: Also, since all you are doing is changing the values and then validating that it either can be set or not and a warning appears or not, then would it be reasonable to put these into one spec with 8 tests?

e2e/regression_guardrail/guardrail.max_blous_max_value.spec.js
```diff
@@ -0,0 +1,16 @@
+const { loop } = require('../../src/index');
```
darinkrauss: [nit] Filename `blous` -> `bolus`? Same for other similar named files.

e2e/regression_guardrail/guardrail.max_blous_max_value.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('guardrail settings max bolus is set to pump maximum of 30 units', () => {
```
darinkrauss: Same comments about testing min/max/warnings from correction range above apply here and anywhere else you are testing guardrails.

e2e/regression_guardrail/guardrail.correction_range_max_warning.spec.js
darinkrauss: BTW, if the reason you aren't checking that a warning is displayed is because it isn't actually implemented in Tidepool Loop yet, then I definitely recommend a TODO comment where the test code should be.

e2e/regression_guardrail/guardrail.max_blous_min_value.spec.js
```diff
@@ -0,0 +1,16 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail settings max bolus minimum allowed is 0 units', () => {
```
darinkrauss: [nit] Are we intending on allowing 0 maximum bolus? Are we supporting the use case of someone *only* allowing Tidepool Loop to manage basals with no bolusing allowed? If so, then setting maximum bolus to 0 seems like an odd way to do that. I would imagine a more direct and obvious way to allow that would be to have a "Allow bolusing" on-off toggle. This may not be a question you know the answer to, but if you do, I'm interested.

e2e/regression_guardrail/guardrail.max_temp_basal_over_limit.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('max temp basal for corrections is either pump maximum or 35 u/hr if the pump doesn’t have a maximum', () => {
+    it('launch loop', async () => {
+        await loop.Launch();
+    });
+    it('open settings add pump simulator, max basal rate set over 35 units', async () => {
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                DeliveryLimits: { maxBolus: '1.0', maxBasalRate: '45.0' }
```
darinkrauss: For anything that checks boundaries for an integer, I would check the very next integer (e.g. check 36 if max is 35). 

For anything that checks boundaries for a float, I would check the very next *tenth* (e.g check 35.1 if max is 35.0; or -0.1 if min is 0.0). This ensures there isn't a bug where the limit is supposed to be 35.0, but the validation is not checking past the decimal point and inadvertently allows 35.1 (perhaps because it rounds first before validating).

e2e/regression_guardrail/guardrail.scheduled_basal_rate_max_basal_rate.spec.js
```diff
@@ -0,0 +1,17 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail settings basal rate cannot be higher than the max basal rate', () => {
+    it('launch loop', async () => {
+        await loop.Launch();
+    });
+    it('setup without Insulin Sensitivities applied', async () => {
```
darinkrauss, line 7: [nit] Name of test seems to be incorrect?

e2e/regression_risk/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,30 @@
+const { loop } = require('../../src/index');
+
+describe.skip('Closed loop is stopped when we have cgm data loss', () => {
+    let config = {
+        scenario: 'flat_cgm_trace',
+        settings: loop.settings.default,
+    };
+    afterAll(async () => {
+        await loop.RemoveData();
```
darinkrauss, line 9: If we are restarting the simulator and reinstalling the app for each spec, then do we really need to remove data `afterAll`?

e2e/regression_risk/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        await loop.AdvanceScenario(config.scenario, '1');
+    });
+    it('should have no status alert', async () => {
+        await loop.screens.home.ExpectNoLoopStatusAlert();
+    });
+    it('cgm data is turned off', async () => {
+        await loop.screens.settings.SetCGMSimulatorSettings({ modelData: { bgValues: [], model: loop.settings.cgmModel.None } })
+    });
+    it('should have CGM data status alert', async () => {
+        await loop.screens.home.ExpectLoopStatusGlucoseDataAlert();
```
darinkrauss, line 28: Is everything else setup correctly here to isolate the missing CGM data as the only error?

e2e/regression_risk/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,30 @@
+const { loop } = require('../../src/index');
```
darinkrauss, line 1: [nit] Consider renaming this test to indicate it is about **cgm** data loss.

e2e/regression_settings/settings.correction_range_below_suspend_threshold.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    it('should apply settings', async () => {
+        let applySettings = loop.settings.default;
+        applySettings.CorrectionRanges = [{ time: '12:00 AM', min: '120', max: '145' }];
+        applySettings.SuspendThreshold = { value: '150' };
+        await loop.Configure({
+            scenario: 'flat_cgm_trace',
+            settings: applySettings,
+        });
+    });
+    it('should not be in closed loop mode', async () => {
```
darinkrauss, line 19: [nit] This comment is a bit confusing. You might want to consider standardizing on key phrases. For example:

"closed loop mode" - when the user has **enabled** the "Closed Loop" toggle in Settings (makes no indication of a successful loop or not)
"open loop mode" - when the user has **disabled** the "Closed Loop" toggle in Settings
"red loop" - when the user has **enabled** the "Closed Loop" toggle in Settings, but there is an error with last X loops or it hasn't looped
"yellow loop" - when the user has **enabled** the "Closed Loop" toggle in Settings, but there is an error with last Y loops
"green loop" - when the user has **enabled** the "Closed Loop" toggle in Settings, and it was successful the last loop

Or something like that.

In this case, you aren't really testing whether it is is in "closed loop mode" or not, but whether it has successfully looped yet or not, right?

e2e/regression_settings/settings.correction_range_below_suspend_threshold.spec.js
```diff
@@ -0,0 +1,35 @@
...
+        await loop.screens.home.ExpectLoopNotYetRun();
+    });
+    it('should show error that indicates why not in closed loop mode', async () => {
+        //TODO: confirm this??
```
darinkrauss, line 23: TODO?

e2e/regression_settings/settings.correction_range_below_suspend_threshold.spec.js
darinkrauss: Or do I have that wrong?

darinkrauss: If not, then same comment applies to comments in the pack of related files.

e2e/regression_settings/settings.no_insulin_sensitivities.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        await loop.Configure({
+            scenario: 'flat_cgm_trace',
+            settings: loop.settings.filter(loop.settings.default, [loop.settings.type.InsulinSensitivities])
+        });
+    });
+    it('should not be in closed loop mode', async () => {
+        await loop.screens.home.ExpectLoopNotYetRun();
+    });
+    it('should show error that indicates why not in closed loop mode', async () => {
+        await loop.screens.home.ExpectLoopStatusCarbsAlert();
```
darinkrauss, line 17: The inconsistent alerting seems like a bug in Tidepool Loop. The fact that this is complaining about carbs but it is a failure of insulin sensitivities seems like an issue that needs to be addressed in Tidepool Loop. What do you think now that you've done a deep dive on all of the error messaging when certain settings are not properly configured?

e2e/regression_settings/settings.valid.spec.js
```diff
@@ -0,0 +1,26 @@
...
+        await loop.Configure(config);
+    });
+    it('should not be in closed loop mode', async () => {
+        await loop.screens.home.ExpectLoopNotYetRun();
+    });
+    it('should advance the scenario so we are looping', async () => {
+        await loop.AdvanceScenario(config.scenario, '1');
+    });
+    it('should have no status alert', async () => {
+        await loop.screens.home.ExpectNoLoopStatusAlert();
```
darinkrauss: [nit] The `ExpectNoLoopStatusAlert` is a bit misleading here. I initially read this as expect a "no loop" status alert. Perhaps something like "ExpectSuccessfulLoop" to indicate that it is a positive outcome?

e2e/smoke_test/smokeTest.spec.js
```diff
@@ -1,139 +1,126 @@
...
-                await screen.settings.SetClosedLoop();
+                await loop.screens.settings.SetClosedLoop();
             });
             it('can be go set to open loop', async () => {
-                await screen.settings.SetOpenLoop();
+                await loop.screens.settings.SetOpenLoop();
             });
             it('can issue a report', async () => {
-                await screen.settings.IssueReportLabel();
+                loop.screens.settings.IssueReportLabel();
```
darinkrauss: How come no `await` here? 

Or, turn that on its head, do we need the `await` everywhere else assuming that the functions being called (e.g. `SetClosedLoop`, `SetOpenLoop`, `IssueReportLabel`) all call `await` inside of the functions?

src/settingsScreen.js
```diff
@@ -63,6 +63,10 @@ var SettingDefault = {
      *  @summary  ClosedLoop: true
      */
     ClosedLoop: true,
+    /**
+     *  @summary  OpenLoop: true
+     */
+    OpenLoop: false,
```
darinkrauss: [nit] Is this just the opposite of `ClosedLoop`? Might want consider just having `ClosedLoop` be a boolean. (What if the user sets both `ClosedLoop` and `OpenLoop` to true? From the code below `ClosedLoop` wins, but that isn't obvious.)

darinkrauss (approved):
A whole bunch of comments for you to think about. Nothing show stopping, though, and okay to move forward, but you might want to address any of the comments you agree with at some point.

e2e/regression_guardrail/guardrail.correction_range_max_warning.spec.js
jh-bate: yeah according to the gaurdrail docs there is a warning at `120` and also at `70` so your list id correct.

Yeah I will add a todo for the actual warning when it is expected

I can possibly put all eight of these in one test spec but that does depend a little on implementation once it's done. It we redo the pickers also to make them accessible that it should be fine otherwise it is a bit of a nightmare and one test per spec will be cleanest

jh-bate: agreed - have written all in one spec file with a todo for warning and also to refactor once it is simple to located and keep updating the pickers

e2e/regression_guardrail/guardrail.max_blous_max_value.spec.js
jh-bate: agreed - have updated these also to match the correction range test pattern

e2e/regression_settings/settings.no_insulin_sensitivities.spec.js
jh-bate: yeah good point - I need to chat with becky and raise bugs where appropriate as they don't make sense in all instances

e2e/regression_settings/settings.valid.spec.js
jh-bate: good idea

src/settingsScreen.js
jh-bate: updated

e2e/smoke_test/smokeTest.spec.js
jh-bate: I was mistakenly just checking for the presence of the label rather than interacting with it and opening the section. I have updated the code and test 

I have played around with the awaits and they are required when interacting, they aren't required when selecting (i.e. getting a certain button element)

PR #16, opened 2020-04-08
LOOP-987 LOOP-988 Jhbate/sprint14

jh-bate:
@darinkrauss just wanted a quick sanity check on the direction of this work. The thing i need to do for this test is check that there is an alert when trying to set something, todo this I am passing an optional function to do these checks.

Opinions??

### DIFF
--- a/e2e/regression_settings/settings.no_max_basal.spec.js
+++ b/e2e/regression_settings/settings.no_max_basal.spec.js
@@ -1,0 +1,19 @@
+const { loop } = require('../../src/index');
+
+describe('Closed loop is not allowed when settings are not applied for max basal', () => {
+    it('launch loop', async () => {
+        await loop.Launch();
+    });
+    it('setup without max basal applied', async () => {
+        let noMaxBasalRate = loop.settings.default;
+        noMaxBasalRate.DeliveryLimits = { maxBolus: '10.0', maxBasalRate: '' };
+        await loop.Configure({ scenario: 'flat_cgm_trace', settings: noMaxBasalRate });
+    });
+    it('should not be in closed loop mode', async () => {
+        await loop.screens.home.ExpectLoopNotYetRun();
+    });
+    it('should show error that indicates why not in closed loop mode', async () => {
+        await loop.screens.home.ExpectLoopStatusInsulinAlert()
+    });
+});
+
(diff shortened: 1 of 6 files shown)


e2e/regression_settings/settings.no_max_basal.spec.js
```diff
@@ -0,0 +1,17 @@
+const { loop } = require('../../src/index');
+
+describe('Closed loop is not allowed when settings are not applied for max basal', () => {
+    it('launch loop', async () => {
+        await loop.Launch();
```
darinkrauss, line 5: [nit] If all of these functions being called (e.g. `loop.Launch()`, `loop.screens.settings.Open`) internally use `await`, is it possible and/or a good idea to get rid of the `await` at this level? Is it redundant?

e2e/regression_settings/settings.no_max_basal.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    });
+    it('setup without max basal applied', async () => {
+        await loop.screens.settings.Open();
+        await loop.screens.settings.AddPumpSimulator();
+        var checks = async function () {
```
darinkrauss: Is this so you can do these checks before closing the specific settings window? If so, then this makes sense, but I might call them `expectations` rather than `checks` (since Detox uses "expect" terminology).

src/settingsScreen.js
```diff
@@ -364,10 +364,11 @@ class SettingsScreen {
...
     }
     /**
-     * @param {object} { maxBasalRate string, maxBolus string }
-     * @example await settings.SetDeliveryLimits({maxBasalRate:'1.0', maxBolus:'10.0'})
+     * @param {object} limits { maxBasalRate string, maxBolus string }
+     * @param {function} postChecksFn function that is called after the setting is set
+     * @example await settings.SetDeliveryLimits({maxBasalRate:'1.0', maxBolus:'10.0'}, checks)
      */
-    async SetDeliveryLimits(limits) {
+    async SetDeliveryLimits(limits, postChecksFn) {
```
darinkrauss: [nit] I would rename `postChecksFn` to `additionalExpectations`.

e2e/regression_settings/settings.no_max_basal.spec.js
darinkrauss: In looking through some of the code, I see multiple levels of `await` which I don't think are necessary. It would be good to come up with a common process on where `await` is called and where it is not.

darinkrauss (approved):
A couple nits and a suggestion about the use of `await`. Otherwise, the direction you are going LGTM! 🎸

src/settingsScreen.js
jh-bate: good idea

e2e/regression_settings/settings.no_max_basal.spec.js
jh-bate: agreed as discussed they are required unfortunately

PR #17, opened 2020-04-13
LOOP-1065 guardrail settings > Insulin Sensitivity Factor

jh-bate:
https://tidepool.atlassian.net/browse/LOOP-1065

### DIFF
--- a/e2e/regression_guardrail/guardrail.insulin_sensitivities.spec.js
+++ b/e2e/regression_guardrail/guardrail.insulin_sensitivities.spec.js
@@ -1,0 +1,54 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail insulin sensitivities', () => {
+    it('cannot set 501 units', async () => {
+        await loop.Launch();
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '501' }],
+            }
+        });
+    });
+    it('set 500 units, warning', async () => {
+        await loop.Launch();
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }],
+            }
+        });
+        //TODO assert on warning
+    });
+    it('set 400 units, warning', async () => {
+        await loop.Launch();
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '400' }],
+            }
+        });
+        //TODO assert on warning
+    });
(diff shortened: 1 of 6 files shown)


e2e/regression_guardrail/guardrail.insulin_sensitivities.spec.js
```diff
@@ -0,0 +1,65 @@
+const { loop } = require('../../src/index');
+
+describe('guardrail insulin sensitivities', () => {
+    it('cannot set 501 units', async () => {
+        await loop.Launch();
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '501' }],
```
darinkrauss: [nit] Since I see `loop.Launch` and `loop.Configure` as test "setup", consider breaking the actual test out separately from the `loop.Configure` code, if possible.

e2e/regression_guardrail/guardrail.insulin_sensitivities.spec.js
```diff
@@ -0,0 +1,65 @@
...
+        await loop.Launch();
+        await loop.Configure({
+            settings: {
+                AddPumpSimulator: true,
+                InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '400' }],
+            }
+        });
+        //TODO assert on warning
+    });
+    it('set 15 units, warning', async () => {
```
darinkrauss, line 18: [nit] I would only have a pair of tests after every "boundary". So, 501 (to show it cannot be set) and 500 (to show that it can be set with a warning). Then, if one type of alert is given for 400+, check 400 and 399. If another type of alert is given for 200+, check 200 and 199. Etc.. A "boundary" is where the fundamental behavior changes. Does that make sense?

src/settingsScreen.js
```diff
@@ -434,7 +434,11 @@ class SettingsScreen {
             for (let index = 0; index < sensitivities.length; index++) {
                 const sensitivity = sensitivities[index];
                 await match.accessible.ButtonBarButton(text.general.Add).tap();
-                await match.accessible.Label(`${sensitivity.bgValuePerInsulinUnit} ${unitsSuffix}`).atIndex(1).tap();
+                let currentMax = 500;
```
darinkrauss: [nit] Consider moving this constant to a "loop constants" file so as these things change there is one place for them all to update.

darinkrauss (approved):
A few nits, but otherwise LGTM! 🚢

src/settingsScreen.js
jh-bate: good idea

e2e/regression_guardrail/guardrail.insulin_sensitivities.spec.js
jh-bate: yeah good plan

PR #19, opened 2020-04-19
LOOP-1107 Jhbate/test init

jh-bate:
reorganisation

 - now either `smoke` or `regression` test

 - test setup is more configurable allowing the passing of 
    - scenario name
    - settings
    - start screen (start on the screen you want for the test)
    - filter (for settings)
    - language (currently a noop but used in the future) 
    - target (DIY or Tidepool - currently a noop but used in the future) 

 - language is now passed into screens 

- test setup hides more of the internals so the test creator doesn't need to worry 

example:

```
loopTest = await new LoopTest.Builder(target.tidepool)
            .withScenario('flat_cgm_trace')
            .withSettings(setting.default)
            .withSettingsFilter([setting.type.BasalRates])
            .withStartScreen(screenName.settings)
            .build();
```

the setup is now done here https://github.com/tidepool-org/LoopUITests/blob/jhbate/test-init/src/loopTest.js


e2e/regression/guardrail.correction_range.spec.js
```diff
@@ -0,0 +1,50 @@
+const { LoopTest, screenName, target } = require('../../src/index');
```
darinkrauss: [nit] Perhaps `LoopTest` -> `Test`?

e2e/regression/guardrail.correction_range.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+describe.skip('guard rail correction range', () => {
```
darinkrauss: [nit] Just making note this is `skip`.

e2e/regression/guardrail.correction_range.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    var loopTest;
+    it('should setup with correct configuration', async () => {
+        loopTest = await new LoopTest.Builder(target.tidepool)
```
darinkrauss: [nit] Consider something like:
```
test = new Test().withSettings({ AddPumpSimulator: true})
                              .withStartScreen(screenNames.settings);
await test.prepare();
```
Note: 
  - `loopTest` -> `test`
  - `LoopTest` -> `Test`
  - no separate `Builder` object
  - move `target.tidepool` to top-level test-run construction so it can be configured outside of the test framework
  - `build` -> `prepare` (as in preparing the test)

e2e/regression/guardrail.correction_range.spec.js
```diff
@@ -0,0 +1,50 @@
...
+            .withSettings({ AddPumpSimulator: true })
+            .withStartScreen(screenName.settings)
+            .build();
+    });
+
+    // TODO: once CorrectionRanges uses accessibile pickers we should
+    // be able to launch loop once and then reuse the setup
+    it('maximum 180 can be set, warning', async () => {
+        await loopTest.settingsScreen.SetCorrectionRanges([{ time: '12:00 AM', min: '179', max: '180' }]);
```
darinkrauss: Any thoughts on how to represent these `mg/dL` values in `mmol/L` units for tests when (DIY|Tidepool) Loop are set to `mmol/L`? Consider putting these values in something separate that can be configured for `mg/dL` or `mmol/L` mode. Perhaps something like `limits.correctionRange.max.maximum`, `limits.correctionRange.max.warning`.

e2e/regression/guardrail.correction_range.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        //TODO assert on warning
+    });
+    it('maximum 120 can be set, no warning', async () => {
+        await loopTest.settingsScreen.SetCorrectionRanges([{ time: '1:00 AM', min: '119', max: '120' }]);
+    });
+    it('minimum 70 can be set, no warning', async () => {
+        await loopTest.settingsScreen.SetCorrectionRanges([{ time: '1:30 AM', min: '70', max: '180' }]);
+    });
+    it('minimum 69 can be set, warning', async () => {
+        await loopTest.settingsScreen.SetCorrectionRanges([{ time: '2:00 AM', min: '69', max: '180' }]);
```
darinkrauss: Does this case actually display two warnings? One for the min < 70 and one for max > 120? If so, might want that in the TODO so you remember.

darinkrauss: Actually, I would follow what you've done for the first three tests. Specifically, 179/180, 120/121, 119/120. 

Continue that in these tests.  Specifically, 69/70, 60/61.

Then add new tests to test warnings on both min and max. Specifically 69/121.

Finally, just for the sake of reading, I'd put the 181 test at the top so all of the tests are in order from maximum max to minimum min.

e2e/regression/guardrail.insulin_sensitivities.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    });
+
+    it('set 500 units, warning', async () => {
+        await loopTest.settingsScreen.SetInsulinSensitivities([{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }]);
+        //TODO assert on warning
+    });
+    it('set 400 units, warning', async () => {
+        await loopTest.settingsScreen.SetInsulinSensitivities([{ time: '12:30 AM', bgValuePerInsulinUnit: '400' }]);
+        //TODO assert on warning
+    });
```
darinkrauss, line 25: What about tests for 399 and 16 (or whatever are the values that just barely don't give you a warning on the high and low sides)?

e2e/regression/guardrail.max_ bolus.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        //TODO assert on warning
+    });
+    it('set 30 units, warning', async () => {
+        await loopTest.settingsScreen.SetDeliveryLimits({ maxBolus: '30.0', maxBasalRate: '1.0' });
+        //TODO assert on warning
+    });
+    it('set 20 units, warning', async () => {
+        await loopTest.settingsScreen.SetDeliveryLimits({ maxBolus: '20.0', maxBasalRate: '1.0' });
+        //TODO assert on warning
+    });
```
darinkrauss, line 32: Tests for values that don't have warnings? 19.9 and 0.1 maybe?

e2e/regression/guardrail.insulin_sensitivities.spec.js
darinkrauss: Basically, I think when testing boundary conditions you want to test both sides of the boundary.

e2e/regression/guardrail.max_temp_basal.spec.js
```diff
@@ -0,0 +1,21 @@
...
+            .withSettings({ AddPumpSimulator: true })
+            .withStartScreen(screenName.settings)
+            .build();
+    });
+
+    it('set 35 units', async () => {
+        await loopTest.settingsScreen.SetDeliveryLimits({ maxBolus: '1.0', maxBasalRate: '35.0' });
+        //TODO assert on warning
+    });
+    it('cannot set 36 units', async () => {
```
darinkrauss: [nit] 35.1 instead of 36.0 assuming the max basal rate goes up in 0.1 increments?

e2e/regression/guardrail.max_temp_basal.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        await loopTest.settingsScreen.SetDeliveryLimits({ maxBolus: '1.0', maxBasalRate: '36.0' });
+        //TODO assert warning
+    });
```
darinkrauss, line 32: The rest of the boundaries? With warning high, without warning high, without warning log, with warning low, cannot set low?

e2e/skip/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,31 @@
+const { loop, setting, screenName, target } = require('../../src/index');
+
+describe.skip('Closed loop is stopped when we have cgm data loss', () => {
+
+    var loopTest;
+    afterAll(async () => {
+        await loopTest.removeData();
```
darinkrauss: If the simulator is going to be destroyed after this, then why bother with remove data?

e2e/skip/cgm.data_loss.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+    it('should setup with correct configuration', async () => {
+        loopTest = await new LoopTest.Builder(target.tidepool)
+            .withScenario('flat_cgm_trace')
+            .withSettings(setting.default)
+            .withStartScreen(screenName.home)
+            .build();
+    });
+    it('should not be in closed loop mode', async () => {
+        await loopTest.homeScreen.ExpectLoopNotYetRun();
```
darinkrauss: How does `loopTest.homeScreen` interact with the `screenName` object? Not sure how they are related.

e2e/skip/settings.correction_range_below_suspend_threshold.spec.js
```diff
@@ -0,0 +1,37 @@
...
+        await loopTest.homeScreen.ExpectLoopNotYetRun();
+    });
+    it('should show error that indicates why not in closed loop mode', async () => {
+        //TODO: confirm this??
+        await loopTest.homeScreen.ExpectLoopStatusCarbsAlert();
+    });
+    it('should only enter carbs', async () => {
+        await loopTest.carbEntryScreen.Open();
+        await loopTest.carbEntryScreen.SetCarbs('30');
+    });
```
darinkrauss, line 30: Is there anything else we are verifying here other than being able to set 30 carbs?

darinkrauss: Perhaps this test should be combine with the one below so you have one test that enters 30 carbs and saves without bolusing?

e2e/skip/settings.no_insulin_sensitivities.spec.js
```diff
@@ -0,0 +1,20 @@
+const { LoopTest, setting, target, screenName } = require('../../src/index');
+
+describe('Closed loop is not allowed when settings are not applied for Insulin Sensitivities', () => {
+    var loopTest;
+    it('should without Insulin Sensitivities applied', async () => {
+        loopTest = await new LoopTest.Builder(target.tidepool)
+            .withScenario('flat_cgm_trace')
+            .withSettings(setting.default)
+            .withSettingsFilter([setting.type.InsulinSensitivities])
```
darinkrauss, line 9: Consider `.withSettingsFilter([setting.type.InsulinSensitivities])` -> `.withoutSettings(setting.type.InsulinSensitivities, ...)`?

e2e/skip/settings.no_max_bolus.spec.js
```diff
@@ -0,0 +1,19 @@
+const { LoopTest, target, screenName } = require('../../src/index');
+
+describe('Closed loop is not allowed when settings are not applied for max bolus', () => {
+    var loopTest;
+    it('should lauch with pump simulator', async () => {
```
darinkrauss: [nit] `lauch` -> `launch`

scripts/test.sh
```diff
@@ -14,10 +14,11 @@ TEST_DIRECTORY="${SCRIPT_DIRECTORY}/.."
 
 error() {
   echo "ERROR: ${*}" >&2
-  echo "Usage: ${SCRIPT} <build-root> <configuration>" >&2
+  echo "Usage: ${SCRIPT} <build-root> <configuration> <type>" >&2
   echo "Parameters:" >&2
   echo "  <build-root>      root of the build that contains the app" >&2
   echo "  <configuration>   detox configuration to use" >&2
+  echo "  <type>            type of tests to run" >&2
```
darinkrauss: Might want to consider listing the potential options here?

scripts/test.sh
```diff
@@ -33,6 +34,8 @@ BUILD_ROOT="${1}"
 shift 1
 CONFIGURATION="${1}"
 shift 1
+TEST_TYPE="${1}"
```
darinkrauss, line 37: I actually don't see where this is used. Future work?

src/properties.js
```diff
@@ -0,0 +1,137 @@
...
+    None: 'No Data'
+};
+
+/**
+ * @summary Defaults that can be used to apply to all settings
+ * @example await settings.Apply(SettingDefault)
+ */
+var settingDefault = {
+    /**
+     * @summary DeliveryLimits: { maxBolus: '10.0', maxBasalRate: '3.0' }
```
darinkrauss: Are you expected to have generated documentation? If not, don't bother with these comments as they are redundant and will likely just go stale over time.

src/properties.js
```diff
@@ -0,0 +1,137 @@
...
+     */
+    InsulinModel: settingInsulinModel.RapidChildren,
+    /**
+     * @summary CarbRatios: [{ time: '12:00 AM', carbGramsPerInsulinUnit: '8' }]
+     */
+    CarbRatios: [{ time: '12:00 AM', carbGramsPerInsulinUnit: '8' }],
+    /**
+     * @summary InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }]
+     */
+    InsulinSensitivities: [{ time: '12:00 AM', bgValuePerInsulinUnit: '500' }],
```
darinkrauss, line 47: For all of these settings, consider using values that might be reasonable for an average PwD (however you define "average").

darinkrauss (approved):
A whole pile of nits (even if I got lazy and didn't put "[nit]" in each comment they are all nits). A good step in the right direction. LGTM! 🇳🇿

src/properties.js
jh-bate: Yeah I spoke with Shawn about this, for the risk assessment they have a 'average' users settings that I plan to use

PR #20, opened 2020-04-29
LOOP-1107 LOOP-1066 LOOP-1064 re-org for running CI tests efficiently 

jh-bate:
- updated `limits` for guard-rails tests 

- single `smoke` and `regression` test that allow to run multiple tests that are composed within
  - provides a massive speed up as the simulator setup is only done once 
  - allows great flexibility as we can import focused tests groups to be run as part of a larger test 

- tests by area are bundled and then imported to run as part of an overall test in a simulator

- settings broken into sub screens
  - sub screen returned on `Open` function
  - once sub screen is opened you can perform its actions  (Apply,Save,Cancel etc)


e2e/guardrails/index.js
```diff
@@ -0,0 +1,19 @@
+const { maxBolus } = require('./max_bolus');
+const { basalRateSchedules } = require('./basal_rate_schedules');
+const { insulinSensitivities } = require('./insulin_sensitivities');
+const { maxTempBasalRate } = require('./max_temp_basal_rate');
+const { suspendThreshold } = require('./suspend_threshold');
+const { correctionRange } = require('./correction_range');
+
+var guardrailsTests = {
+    maxBolus,
```
darinkrauss: I would call each of these `Tests`, as well, to avoid confusion. So, `maxBolusTests`, `basalRateScheduleTests`, `insulinSensitivityTests`, etc.

Also, note the lack of plurals on what is being tested. Also, might want to consider using "Schedule" on everything that is a schedule. Specifically, "basalRateScheduleTests", "insulinSensitivityScheduleTests", "correctionRangeScheduleTests", "carbohydrateRatioScheduleTests", etc.

darinkrauss: Also, I think the filename could use the ".tests.js", as well. Perhaps you could use that to indicate these are individual tests that cannot be run independently, as opposed to a spec?

e2e/guardrails/suspend_threshold.js
```diff
@@ -0,0 +1,60 @@
...
+        );
+        //TODO assert on warning
+    });
+    it('can set min units with warning', async () => {
+        await test.settingsScreen.SetSuspendThreshold(
+            { value: limits.suspendThreshold.min.warning }
+        );
+        //TODO assert on warning
+    });
+    it('cannot set above max units', async () => {
```
darinkrauss: [nit] I would order all of these tests from the highest setting to the lowest setting. So, in short, I'd move "above max units" to the top of this file.

Also, I don't see a test for `limits.suspendThreshold.min.limit` here.

src/limits.js
```diff
@@ -0,0 +1,68 @@
...
+    basalDelivery: {
+        step: 0.1,
+        max: {
+            limit: 35,
+        },
+    },
+    basalRates: {
+        step: 0.05,
+        max: {
+            limit: 35,
```
darinkrauss: [nit] Consistent use of `.0`?

e2e/guardrails/basal_rate_schedules.js
```diff
@@ -0,0 +1,61 @@
+const { limits } = require('../../src/index');
+
+var basalRateSchedules = (test) => {
+    it('can set max limit', async () => {
+        await test.settingsScreen.SetBasalRates(
+            [{ time: '12:00 AM', unitsPerHour: limits.basalRates.max.limit }],
+            { fromSettings: true, toSettings: false }
```
darinkrauss: What is the "fromSettings" and "toSettings"?

e2e/guardrails/basal_rate_schedules.js
```diff
@@ -0,0 +1,61 @@
...
+        );
+        //TODO assert on warning
+    });
+    it('can set max warning', async () => {
+        await test.settingsScreen.SetBasalRates(
+            [{ time: '12:30 AM', unitsPerHour: limits.basalRates.max.warning - limits.basalRates.step }],
```
darinkrauss: [nit] Now I am unclear on what `SetBasalRates` actually does? Does it *set* the basal rates to the specified schedule, or does it *add* to the already existing basal rate schedule?

Also, might want to consider `SetBasalRateSchedule`.

darinkrauss (commented):
I added a number of comments for you to consider, but I did not go through with the intent to approve. I like the way this is going, but there may be a few things to iron out.

e2e/guardrails/index.js
jh-bate: these should all be addressed now.

I have moved all the test files into a `tests` dir and also named the files that way too

e2e/guardrails/suspend_threshold.js
jh-bate: I have tried to order all the tests like this now. There was an issue when trying to apply a setting that was out of bounds which is why they were always last

e2e/guardrails/basal_rate_schedules.js
jh-bate: removed. Now with the returning of sub screens I can do away with all this

src/limits.js
```diff
@@ -0,0 +1,68 @@
...
+            warning: 20.0,
+        },
+        min: {
+            limit: 0.0,
+        },
+    },
+    basalDelivery: {
+        step: 0.1,
+        max: {
+            limit: 35,
```
darinkrauss: No `min.limit` here? 0?

src/limits.js
```diff
@@ -0,0 +1,68 @@
...
+var limits = {
+    correctionRange: {
+        step: 1,
+        max: {
+            limit: 180,
+            warning: 120,
+        },
+        min: {
+            limit: 60,
+            warning: 70,
```
darinkrauss, line 12: [nit] Consider on `min` to do `warning` first and then `limit`. It makes it more readable overall. (That is, it is easier to read `180`, `120`, `70`, `60`.)

src/match.js
```diff
@@ -137,7 +126,7 @@ const match = {
             const currentPicker = 0;
             return element(
                 by.type('UIPickerView')
-            ).atIndex(currentPicker).setColumnToValue(column, strValue)
+            ).atIndex(currentPicker).setColumnToValue(column, String(strValue))
```
darinkrauss: [nit] Isn't `String(strValue)` redundant? Perhaps rename `strValue` to just `value`?

src/test.js
```diff
@@ -90,6 +90,11 @@ class Test {
         if (!this.language) {
             this.language = text;
         }
+        if (!this.startScreen) {
+            if (this.settings || this.simulators) {
+                this.startScreen = screenName.home;
+            }
```
darinkrauss, line 96: What will happen if there is no start screen specified, but settings and simulators are missing?

tests/guardrails/insulin_sensitivity_schedule.tests.js
```diff
@@ -0,0 +1,60 @@
+const { limits } = require('../../src/index');
+
+var insulinSensitivityScheduleTests = (test) => {
+    var screen;
+    it('open', async () => {
+        screen = await test.settingsScreen.OpenInsulinSensitivitiesScreen();
+    });
+    it('cannot set above max limit', async () => {
+        try {
+            await screen.Apply({ time: '12:00 AM', bgValuePerInsulinUnit: limits.insulinSensitivities.max.limit + limits.insulinSensitivities.step });
```
darinkrauss, line 10: I love this new `screen` object with the very generic `Apply`, `Save`, and `Edit` functions.

tests/guardrails/insulin_sensitivity_schedule.tests.js
```diff
@@ -0,0 +1,60 @@
...
+        } catch (error) {
+            //TODO assert cannot be set
+            await screen.Save();
+        }
+    });
+    it('can set max limit', async () => {
+        await screen.Edit({ time: '12:00 AM', bgValuePerInsulinUnit: limits.insulinSensitivities.max.limit });
+        await screen.Save();
```
darinkrauss, line 18: [nit] Do you actually have to save the screen for it to warn?

tests/guardrails/insulin_sensitivity_schedule.tests.js
```diff
@@ -0,0 +1,60 @@
...
+    it('cannot set above max limit', async () => {
+        try {
+            await screen.Apply({ time: '12:00 AM', bgValuePerInsulinUnit: limits.insulinSensitivities.max.limit + limits.insulinSensitivities.step });
+        } catch (error) {
+            //TODO assert cannot be set
+            await screen.Save();
+        }
+    });
+    it('can set max limit', async () => {
+        await screen.Edit({ time: '12:00 AM', bgValuePerInsulinUnit: limits.insulinSensitivities.max.limit });
```
darinkrauss, line 17: What is the difference between `Apply` and `Edit`?

tests/guardrails/max_bolus.tests.js
```diff
@@ -0,0 +1,62 @@
...
+
+var maxBolusTests = (test) => {
+    var screen;
+    it('open', async () => {
+        screen = await test.settingsScreen.OpenDeliveryLimitsScreen();
+    });
+    it('cannot set above the max limit', async () => {
+        await screen.Apply({
+            maxBolus: limits.bolusDelivery.max.limit + limits.bolusDelivery.step,
+            maxBasalRate: '1.0'
```
darinkrauss: [nit] Internally to the `Apply` function can we wrap the `maxBasalRate` value with `String(value)` so this could be an actual number and not a string?

tests/guardrails/max_bolus.tests.js
```diff
@@ -0,0 +1,62 @@
+const { limits } = require('../../src/index');
+
+var maxBolusTests = (test) => {
+    var screen;
+    it('open', async () => {
+        screen = await test.settingsScreen.OpenDeliveryLimitsScreen();
+    });
+    it('cannot set above the max limit', async () => {
+        await screen.Apply({
+            maxBolus: limits.bolusDelivery.max.limit + limits.bolusDelivery.step,
```
darinkrauss, line 11: Does this require a catch since we technically cannot set this value?

tests/guardrails/max_bolus.tests.js
```diff
@@ -0,0 +1,62 @@
...
+    it('can set at min limit', async () => {
+        await screen.Apply({
+            maxBolus: limits.bolusDelivery.min.limit,
+            maxBasalRate: '1.0'
+        });
+        //TODO assert on warning
+    });
+    it('cannot set below the min limit', async () => {
+        await screen.Apply({
+            maxBolus: limits.bolusDelivery.min.limit - limits.bolusDelivery.step,
```
darinkrauss, line 53: Does this require a catch since we technically cannot set this value?

tests/guardrails/max_temp_basal_rate.tests.js
```diff
@@ -0,0 +1,36 @@
+
```
darinkrauss, line 1: [nit] Extra blank line?

tests/guardrails/max_temp_basal_rate.tests.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    it('set at max limit', async () => {
+        await screen.Apply({
+            maxBolus: '1.0',
+            maxBasalRate: limits.basalDelivery.max.limit
+        });
+        //TODO assert on warning
+    });
+    it('cannot above max limit', async () => {
```
darinkrauss, line 25: [nit] Put this test on top so we are consistent with the rest of the test going from maximum to minimum?

tests/guardrails/max_temp_basal_rate.tests.js
```diff
@@ -0,0 +1,36 @@
+
+const { limits } = require('../../src/index');
+
+var maxTempBasalRateTests = (test) => {
```
darinkrauss, line 4: [nit] Suggest you add tests around minimum limit (even it it is 0).

tests/guardrails/suspend_threshold.tests.js
```diff
@@ -0,0 +1,60 @@
+const { limits } = require('../../src/index');
+
+var suspendThresholdTests = (test) => {
+    it('can set max units with warning', async () => {
+        await test.settingsScreen.SetSuspendThreshold(
```
darinkrauss, line 5: I believe the suspend threshold is on a separate screen, so we can mirror all of the other tests with an `OpenSuspendThresholdScreen`, do all of the tests and a `Close`?

tests/guardrails/suspend_threshold.tests.js
```diff
@@ -0,0 +1,60 @@
...
+        );
+        //TODO assert on warning
+    });
+    it('can set min units with warning', async () => {
+        await test.settingsScreen.SetSuspendThreshold(
+            { value: limits.suspendThreshold.min.warning }
+        );
+        //TODO assert on warning
+    });
+    it('cannot set above max units', async () => {
```
darinkrauss, line 40: [nit] At top of file?

tests/guardrails/suspend_threshold.tests.js
```diff
@@ -0,0 +1,60 @@
...
+        );
+        //TODO assert NO warning
+    });
+    it('can set min upper boundary units with warning', async () => {
+        await test.settingsScreen.SetSuspendThreshold(
+            { value: limits.suspendThreshold.min.warning }
+        );
+        //TODO assert on warning
+    });
+    it('can set min units with warning', async () => {
```
darinkrauss, line 34: [nit] Duplicate test?

tests/smoke/accessibility.tests.js
```diff
@@ -0,0 +1,249 @@
...
+        await expect(test.homeScreen.AddMealButton()).toExist();
+    });
+    it('has a Bolus Button', async () => {
+        await expect(test.homeScreen.BolusButton()).toExist();
+    });
+};
+
+var bolusScreenAccessibilityTests = (test) => {
+    it('has to open the bolus screen', async () => {
+        await test.bolusScreen.Open();
```
darinkrauss: [nit] Consider using same mechanism as above with an `OpenBolusScreen` returning a screen object that all of the rest of the tests use.

tests/smoke/accessibility.tests.js
```diff
@@ -0,0 +1,249 @@
...
+    });
+    it('has to close the settings screen', async () => {
+        await test.settingsScreen.Close();
+    });
+};
+
+var settingsCarbRatiosScreenAccessibilityTests = (test) => {
+    var screen;
+    it('open', async () => {
+        await test.settingsScreen.Open();
```
darinkrauss: [nit] Consider even having the settings screen be a separate object. So,
```
settingsScreen = await test.homeScreen.OpenSettingsScreen();
screen = await settingsScreen.OpenCarbRatiosScreen();
```
(Note: `OpenCarbRatiosScreen` -> `OpenCarbRatiosScheduleScreen`?)

tests/smoke/functionality.tests.js
```diff
@@ -0,0 +1,143 @@
...
+            let screen = await test.settingsScreen.OpenCGMSimulatorScreen();
+            await screen.Apply(setting.default.CGMSimulatorSettings);
+            await screen.Close();
+        });
+    });
+    describe('pump', () => {
+        it('can be added', async () => {
+            await test.settingsScreen.AddPumpSimulator();
+        });
+        it.skip('set suspend threshold', async () => {
```
darinkrauss, line 60: [nit] At some point might want to do a global find for all "skip" and fix anything broken.

tests/smoke/index.js
```diff
@@ -0,0 +1,31 @@
+const { homeScreenFunctionalityTests, carbEntryScreenFunctionalityTests, cleanupFunctionalityTests, settingsScreenFunctionalityTests } = require('./functionality.tests');
```
darinkrauss: [nit] Consider splitting onto multiple lines like below.

tests/guardrails/correction_range_schedule.tests.js
```diff
@@ -0,0 +1,111 @@
+const { limits } = require('../../src/index');
+
+var correctionRangeScheduleTests = (test) => {
+    it('cannot set above max', async () => {
```
darinkrauss, line 4: [nit] Can these tests take the same form as above tests where there is one "open" test and then a bunch of individual tests, and then a "close" test?

tests/guardrails/correction_range_schedule.tests.js
```diff
@@ -0,0 +1,111 @@
...
+        });
+        //TODO assert warning
+        await screen.Cancel();
+    });
+    it('can set below max', async () => {
+        var screen = await test.settingsScreen.OpenCorrectionRangeScreen();
+        await screen.Apply({
+            time: '12:00 AM',
+            min: limits.correctionRange.max.limit,
+            max: limits.correctionRange.max.limit - limits.correctionRange.step
```
darinkrauss, line 31: Is this disallowed anyhow because min > max? Perhaps it would be best to test one (either min or max) at a time? Perhaps setting the non-tested one to a valid, non-warning value?

src/settingsScreen.js
```diff
@@ -38,53 +48,336 @@ class SettingsScreen {
...
+        await match.accessible.SetPickerValue(1, `${sensitivity.bgValuePerInsulinUnit} ${config.insulinSensitivitiesUnits}`);
+    }
+}
+
+class CorrectionRangeScreen {
+    constructor(language) {
+        this.language = language;
+    }
+    _pickerIndexForTime(time) {
+        switch (time) {
```
darinkrauss: [nit] Consider just building an array of all 24 times and then doing a find to get the index? Also, should the times be in the "language" file for localization and 24-hour clock?

darinkrauss: Perhaps the array could just be in the language file?

src/settingsScreen.js
```diff
@@ -38,53 +48,336 @@ class SettingsScreen {
...
+            currentMin--;
+        } while (currentMin >= range.min);
+    }
+}
+
+class CarbRatiosScreen {
+    constructor(language) {
+        this.language = language;
+    }
+    CarbRatiosHeader() {
```
darinkrauss: [nit] Isn't this redundant `CarbRatiosScreen.CarbRatiosHeader`?

darinkrauss (approved):
A whole pile of comments, but direction LGTM! ☘️

src/limits.js
jh-bate: true

src/settingsScreen.js
jh-bate: yeah - I will update those

src/settingsScreen.js
jh-bate: true - will move

tests/guardrails/insulin_sensitivity_schedule.tests.js
jh-bate: possibly not - I guess I will know when the work is done in loop

tests/guardrails/insulin_sensitivity_schedule.tests.js
jh-bate: apply will create a new sensitivity, edit will update an existing one

tests/guardrails/max_bolus.tests.js
jh-bate: yeah - detox wants strings but they are numeric values. I will update to keep consistent

tests/guardrails/max_bolus.tests.js
jh-bate: yeah - detox fails as it can't find the value you have asked for

tests/smoke/accessibility.tests.js
jh-bate: yeah it has worked well so I will do the same for the other screens

PR #21, opened 2020-05-11
Cron build

jh-bate:
@darinkrauss it came down to the log level. The tests where running successfully but reported no feedback until they completed. And in the case of these tests that is 18-22 mins - hence the timeout.

I also bumped the dependancy versions just to make sure I was up do date


darinkrauss (approved):
LGTM! 🦙

PR #22, opened 2020-05-14
LOOP-1067 LOOP-1289 suspend threshold + cleanup 

jh-bate:
- have switched to limits setup that no longer has the concept of steps 

```
max: {
   limit: 180,
   warning: 120,
   noWarning: 121,
},
```

- moved to screen model for `SuspendThresholdScreen`

- added action, for setting of screen elements 

- allow interaction with new SwiftUI picker so we can swipe to the max and min values and then move in increments of two where it makes sense ... kind of like a user would


src/limits.js
```diff
@@ -3,12 +3,13 @@
  */
 var limits = {
     correctionRange: {
-        step: 1,
         max: {
             limit: 180,
             warning: 120,
+            noWarning: 121,
```
darinkrauss: Wouldn't `noWarning` be less than `warning` on `max`?

darinkrauss (approved):
LGTM. Everything looked reasonable from what I can tell. Only one question about a `noWarning` value.

src/limits.js
jh-bate: doh .... yes it would! I will adjust

PR #29, opened 2020-06-10
LOOP-1393 Jhbate/test results

jh-bate:
- adding better test output that we can surface in CI build
- adding screen shots, but only when a test fails


darinkrauss (approved):
LGTM! 🔑

PR #31, opened 2020-06-14
LOOP-1357 Jhbate/cgm simulator

jh-bate:
work done so we can start testing for error states within the app. 

I know that I am testing simulators (pump and cgm) but thats more so we know that they work as expected for other tests ... hope that makes sense :)

also cleaned-up some of the final configurable defaults and used the same pattern to load and set them as text and guardrails limits


src/config/mgdLDefaults.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
darinkrauss: [nit] Consider putting these in a `defaults.js` and have, at the top-level, a `mgdL` dictionary and a `mmolL` dictionary. Then you could import it as just `defaults` and use it as `defaults.mgdL`.

src/config/mgdLDefaults.js
```diff
@@ -0,0 +1,16 @@
...
+    insulinSensitivity: {
+        defaultStart: 500
```
darinkrauss: [nit] If this is in the defaults, then using a default prefix here is redundant.

src/subScreen/cgmSimulatorScreen.js
```diff
@@ -6,74 +6,177 @@ class CGMSimulatorScreen {
...
+    RandomHighOutlierEffectLabel() {
+        return match.accessible.Label(this.language.Effect.RandomHighOutlier);
+    }
+    RandomLowOutlierEffectLabel() {
+        return match.accessible.Label(this.language.Effect.RandomLowOutlier);
+    }
+    BackfillGlucoseHistoryLabel() {
+        return match.accessible.Label(this.language.History.BackfillGlucose);
+    }
+
```
darinkrauss, line 51: [nit] extra line?

darinkrauss (approved):
LGTM! 🚅

PR #33, opened 2020-06-17
LOOP-1324  Jhbate/sched basals guardrails

jh-bate:
https://tidepool.atlassian.net/browse/LOOP-1324

### DIFF
--- a/src/config/limits.js
+++ b/src/config/limits.js
@@ -45,14 +45,13 @@ var mgdL = {
         },
     },
     basalRates: {
-        step: 0.05,
         max: {
             limit: 35.0,
-            warning: 20.0,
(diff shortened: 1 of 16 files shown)


src/config/limits.js
```diff
@@ -45,14 +45,13 @@ var mgdL = {
         },
     },
     basalRates: {
-        step: 0.05,
         max: {
             limit: 35.0,
-            warning: 20.0,
```
darinkrauss, line 51: [nit] Why is this going away? Don't we want to test where a warnings starts?

darinkrauss (approved):
LGTM. Only one nit/question. 🚀

PR #36, opened 2020-07-08
LOOP-1358: Pump failures

jh-bate:
- basics for checking of pump failures
   -  note could for some as cannot interact with passcode screen

- screens re-org so we can common base screens and functionality
- language clean


src/base/entriesScreen.js
```diff
@@ -0,0 +1,70 @@
...
+    * @param {object} parentScreen.screenTxt
+    * @param {object} parentScreen.generalTxt
+    * @param {object} config
+    */
+    constructor(parentScreen, config) {
+        super(parentScreen);
+        this.config = config;
+    }
+    InfoLabel() {
+        return match.accessible.Label(this.screenTxt.Info);
```
darinkrauss: [nit] `Txt` -> `Text` (not sure if saving one character is worth it)

src/homeScreen.js
```diff
@@ -1,13 +1,26 @@
...
+    constructor(language, settingsScreenDefaults) {
+
         this.glucoseScreen = new homeSubScreen.GlucoseScreen(language);
         this.activeInsulinScreen = new homeSubScreen.ActiveInsulinScreen(language);
         this.insulinDeliveryScreen = new homeSubScreen.InsulinDeliveryScreen(language);
         this.activeCarbohydratesScreen = new homeSubScreen.ActiveCarbohydratesScreen(language);
+        this.settingsScreen = new SettingsScreen(language, settingsScreenDefaults);
+        this.bolusScreen = new BolusScreen(language);
+        this.carbEntryScreen = new CarbEntryScreen(language);
+        this.header = new homeSubScreen.Header(language);
```
darinkrauss: [nit] What is the `homeSubScreen`? It is not clear from the variable name.

darinkrauss (approved):
A couple of nits, but otherwise LGTM! 🔑

src/base/entriesScreen.js
jh-bate: lol - yeah

src/homeScreen.js
jh-bate: the sub screens that open off the home screen - like the "Active Insulin" section that opens into its own screen when you click it. I could maybe come up with a better name

PR #40, opened 2020-07-30
LOOP-1586: Upgrades part two

jh-bate:
* pull out devices so we can access from both header and settings screen
* common code for devices
* add g6 device so we can test the initial setup page
* add custom presets screen
* add seperate meal bolus screen for carb entry
* tests for hud items
* switching to `.toBeVisible()` to ensure items not only exist but can also be seen


README.md
```diff
@@ -6,23 +6,29 @@ Loop end-to-end automated tests using `detox`
...
 
 All tests
 
-`detox test -c ios.sim.debug`
+`npm run all-test`
 
 ### Smoke test
+Accessibility tests that run through the screen elements making sure the are present
+
+- `npm run smoke1-test`
```
darinkrauss: [nit] I would name all of these targets "test-***". For example, "test-smoke-1", "test-all", etc.

src/base/screen.js
```diff
@@ -4,21 +4,39 @@ const action = require('../action');
...
      */
     constructor(parentScreen) {
-        this.openLabel = parentScreen.openLabel;
-        this.openClickableLabel = parentScreen.openClickableLabel;
-        this.backLabel = parentScreen.backLabel;
         this.screenText = parentScreen.screenText;
         this.generalText = parentScreen.generalText;
+
+        if (parentScreen.open) {
+            if (parentScreen.open.isBtn) {
```
darinkrauss, line 27: [nit] Consider consistently using "Button".

src/base/screen.js
```diff
@@ -27,23 +45,34 @@ class Screen {
...
         }
-        return this.CancelButton();
     }
     ContinueButton() {
         return match.accessible.ButtonBarButton(this.generalText.Continue);
     }
     OpenButton() {
-        if (this.openLabel) {
-            return match.accessible.Button(this.openLabel);
+        if (this.openBtn) {
```
darinkrauss, line 72: [nit] "Button"?

src/bolus/index.js
```diff
@@ -0,0 +1,71 @@
...
+            },
+        });
+    }
+    /**
+     * @override Screen.BackButton()
+     * */
+    BackButton() {
+        return match.accessible.ButtonBarButton(this.backLabel);
+    }
+    async Deliver() {
```
darinkrauss, line 24: [nit] Might want to consider separating out controls versus actions. Perhaps list all of the controls (e.g. "BackButton", "SaveAndDeliverButton") and then list the actions available (e.g. "Deliver", "SetBolusAmount).

src/devices/index.js
```diff
@@ -0,0 +1,63 @@
+const match = require('../match');
+
+const { PumpSimulator } = require('./pumpSimulator');
```
darinkrauss: [nit] Any reason why you aren't just exporting the class directly rather than in an object?

Currently:
```
module.exports = {
    PumpSimulator
};
````
and
```
const { PumpSimulator } = require('./pumpSimulator');
```

versus:
```
module.exports = PumpSimulator;
````
and
```
const PumpSimulator = require('./pumpSimulator');
```

Comment (darinkrauss) [reply src/devices/index.js]: Are there some files that export multiple?

File: src/settings/index.js [hunk 6]
```diff
@@ -1,65 +1,63 @@
-const match = require('./match');
-const { settingsSubScreen } = require('./settingScreen/index');
+const match = require('../match');
+// const { BasalRatesScreen } = require('./basalRatesScreen');
```
darinkrauss, line 2: [nit] Commented out code?

src/devices/pumpSimulator.js
```diff
@@ -2,18 +2,26 @@ const element = require('detox').element;
 const match = require('../match');
 const { base } = require('../base/index');
 
-class PumpSimulatorScreen extends base.Screen {
+class PumpSimulator extends base.Screen {
```
darinkrauss: [nit] Why losing the "Screen" suffix if this still extends "base.Screen"?

darinkrauss (approved):
LGTM! A few nits, but nothing significant. 🎃

src/bolus/index.js
jh-bate: yeah - that would be nice

src/devices/index.js
jh-bate: yeah I will clean those up. Its just a hangover from when I had multiple things in a file

PR #42, opened 2020-08-05
LOOP-1621: allow auth, fix delivery limits interaction

jh-bate:
- add authentication, as a configurable test option,  so that we can deliver a bolus
- add configured pump utility from the test object for tests that require it
- fix delivery limits screen so we can interact with the pickers


src/settings/index.js
```diff
@@ -1,8 +1,8 @@
 const match = require('../match');
 // const { BasalRatesScreen } = require('./basalRatesScreen');
```
darinkrauss: [nit] Still commented out code.

tests/functionality/carbEntryScreen.tests.js
```diff
@@ -1,19 +1,15 @@
-var carbEntryScreen = (test) => {
-    var carbEntryScreen;
+module.exports = carbEntryScreen = (test) => {
```
darinkrauss: [nit] Do you even need the `carbEntryScreen = ` part?

tests/functionality/cgmSimulatorScreen.tests.js
```diff
@@ -1,4 +1,4 @@
-var cgmSimulatorScreen = (test) => {
+module.exports = cgmSimulatorScreen = (test) => {
```
darinkrauss: [nit] Same here. Is `cgmSimulatorScreen = ` necessary? This would apply in all similar cases (and I'll stop commenting on them now).

tests/functionality/index.js
```diff
@@ -1,15 +1,15 @@
-const { settingsScreen } = require('./settingsScreen.tests');
-const { homeScreen } = require('./homeScreen.tests');
-const { carbEntryScreen } = require('./carbEntryScreen.tests');
-const { pumpSimulatorScreen } = require('./pumpSimulatorScreen.tests');
-const { cgmSimulatorScreen } = require('./cgmSimulatorScreen.tests');
-const { deliveryLimitsScreen } = require('./deliveryLimitsScreen.tests');
+const settingsScreen = require('./settingsScreen.tests');
```
darinkrauss: [nit] Consider adding `Tests` suffix on all of these so it is clear you aren't importing the settings screen, but rather the settings screen tests.

darinkrauss (approved):
A couple of nits, but otherwise LGTM! 🐎

tests/functionality/cgmSimulatorScreen.tests.js
jh-bate: good point - have cleaned them all up

PR #43, opened 2020-08-12
LOOP-1783: Allow for authentication during tests

jh-bate:
- added biometric authentication via faceid (can also be turned off at test level)
- added base screen Authenticate method (note had to include a sleep to give the authentication time to return)
- update test script so we can pass through numbered targets e.g. `functional_1` `functional_2` etc
- split functional tests so we can run in parallel


src/bolus/index.js
```diff
@@ -23,13 +23,19 @@ class BolusScreen extends base.Screen {
     }
     async Deliver() {
         await this.DoneButton().tap();
-        await this.SaveAndDeliverButton().tap();
+        await this.DeliverButton().tap();
+    }
+    async EnterBolus() {
+        await this.EnterBolus().tap();
```
darinkrauss: `EnterBolus().tap()` -> `EnterBolusButton().tap()`?

src/test.js
```diff
@@ -105,25 +105,28 @@ class Test {
             }
         }
 
+        this.homeScreen = new HomeScreen(this.language, this.screenDefaults);
+
+        var loopAppPermissions = { notifications: 'YES', health: 'YES', };
```
darinkrauss: [nit] Trailing comma?

darinkrauss (approved):
One thing I think you'll need to fix, but otherwise LGTM! 🧑‍🚀

PR #45, opened 2020-08-18
LOOP-1782: New settings

jh-bate:
- drive settings from therapy screen
- starting to removing old settings from tests


darinkrauss (approved):
LGTM! 🚀

PR #47, opened 2020-08-30
AUTO-155: create config for iPhone 6, iPhone SE (2nd generation) and iPhone 11 Pro

jh-bate:
- smoke tests (not all tests) now setup so we can run against different device sizes
- build updated to match ci build
- ability to now load mock therapy settings


README.md
```diff
@@ -14,6 +14,12 @@ All tests
 
 `npm run test-all`
 
+## Device Configurations
+
+- iPhone 11 Pro `ios.sim.debug-11pro`
+- iPhone 6 `ios.sim.debug-6`
```
darinkrauss, line 20: [nit] Consider more explicit naming (as "6" could be anything). For example "ios.sim.debug.iphone-11-pro", "ios.sim.debug.iphone-6", "ios.sim.debug.iphone-se-2".

darinkrauss (approved):
Just one nit for your consideration, but otherwise LGTM! 🚓

README.md
jh-bate: I was thinking `ios` implied `iphone` hence i didn't add that. Thoughts??

darinkrauss: Ah, I suppose. I was interpreting that was "ios.sim" - iOS Simulator, "debug" - Debug build, and everything else was the device. Up to you, but "6" could be interpreted as a version or a device type.

jh-bate: makes sense - have updated in the next PR. Thanks

PR #50, opened 2020-09-03
AUTO-187, AUTO-182: Mocked therapy settings

jh-bate:
A big update for the new therapy settings work key points are 

- now uses "mocked therapy settings" as the start point 
- guardrails test language up dated to be consistent 
- guardrails tests broken out to test each part separately. Setting of the value, presence (or absence) of  the warning icon and message
- rework test setup so all values are options passed in. Can now also start in closed loop mode with cgm data 
- general cleanup


src/config/limits.js
```diff
@@ -1,3 +1,4 @@
+//update based on therapy settings
 var mgdL = {
```
darinkrauss, line 2: FYI - There are tickets coming through that will make the therapy setting warnings and limits different based upon the age (which is calculated from entered birthdate). Something to think about before it happense.

src/settings/suspendThresholdScreen.js
```diff
@@ -30,9 +30,9 @@ class SuspendThresholdScreen extends base.EntryScreen {
     Header() {
         return match.accessible.TextLabel(this.screenText.Header).atIndex(0);
     }
-    async OpenPicker() {
-        await match.accessible.TextLabel(this.bgUnitsLabel).atIndex(0).tap();
-    }
+    // async OpenPicker() {
```
darinkrauss: [nit] Commented out? Can it be removed?

tests/guardrails/basalRateSchedule.tests.js
```diff
@@ -1,83 +1,59 @@
+const name = require('./testNames');
+
 module.exports = (test) => {
     var screen;
-    var settings;
+    var therapyScreen;
```
darinkrauss: [nit] `therapySettingsScreen`?

tests/guardrails/correctionRangeSchedule.tests.js
```diff
@@ -1,122 +1,97 @@
+const name = require('./testNames');
+
 module.exports = (test) => {
     var screen;
-    var settings;
+    var therapyScreen;
```
darinkrauss: [nit] `therapySettingsScreen`? (Last message about this. Assume this applies to all instances throughout code.)

tests/guardrails/suspendThreshold.tests.js
```diff
@@ -1,90 +1,74 @@
...
+                current: { value: screenLimit.min.limit },
             });
         });
-        it('check for no guardrail warning icon', async () => {
-            await expect(screen.GuardrailWarningIconPicker()).toBeNotVisible();
+        it(name.HasGuardrailIcon, async () => {
+            await expect(screen.GuardrailWarningIconPicker()).toBeVisible();
+        });
+        it(name.HasGuardrailMessage, async () => {
+            await expect(screen.GuardrailMessage('Low Suspend Threshold')).toBeVisible();
```
darinkrauss: [nit] Is this really the message "Low Suspend Threshold" or is this a key that maps to the actual message. If the latter, then consider wrapping like we do all other strings. So, something like `screen.LowSuspendThresholdGuardrailMessage()`?

tests/guardrails/testNames.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
```
darinkrauss: What is this used for?

darinkrauss (approved):
LGTM! There are a couple things you might want to address, but I see no issues. 🎃

tests/guardrails/testNames.js
jh-bate: just to keep the naming consistent across all of the guardrails tests ... I would change the naming in one and then it be incorrect in the others

src/config/limits.js
jh-bate: yeah - I am mulling over this in general and was going to have a chat to you. I need dynamic setup based on the therapy settings

PR #53, opened 2020-10-05
iphoneSE2 test updates

jh-bate:
changes to allow for the smaller screen size of the iPhone SE, and also use of fingerid rather than faceid for authentication.

Biggest change is probably in  `SwipeDownUntil(desiredLabel)` which will keep swiping until the desired label is found on the screen. This was because of the different screen sizes more swipes were required to see the label we wanted. I have added a get out clause to the function so it doesn't end out in an infinite loop trying to find a label that doesn't exist


src/action.js
```diff
@@ -16,6 +16,25 @@ var _nextPickerStep = function (currentValue, expectedValue) {
...
+    let labelNotSeen = true;
+    let timesSwiped = 0;
+    const maxSwips = 15;
+    do {
+        if (timesSwiped >= maxSwips) {
+            throw `Swiped screen ${direction} maximum times, check setup as there may be an issue`;
+        }
+        try {
+            await expect(desiredLabel).toBeVisible();
+            labelNotSeen = false;
```
darinkrauss: [nit] The `labelNotSeen` was hard for me to reason about. Perhaps reverse the logic so the variable is named `labelVisible`, with initial state of `false`, it is set to `true` at this point, and the do/while loop is while `!labelVisible`. Double-negatives are usually harder to reason about.

src/base/screen.js
```diff
@@ -69,7 +83,11 @@ class Screen {
         return match.accessible.Button(this.generalText.Save);
     }
     async Authenticate() {
-        await device.matchFace();
+        if (_deviceInfo().useFaceID) {
```
darinkrauss: [nit] Is there an instance of `Screen` that can hold an instance of `deviceInfo` so it doesn't need to be created every go around?

src/settings/therapyScreen.js
```diff
@@ -75,7 +74,7 @@ module.exports = class TherapyScreen extends base.Screen {
         return this.premealRangeScreen.InfoLabel();
     }
     async OpenPreMealRangeScreen() {
-        await this.SwipeUp(this.PreMealRangeLabel());
+        await this.SwipeUpUntil(this.PreMealRangeLabel());
```
darinkrauss: [nit] Consider renaming `SwipeUpUntil` to `SwipeUpUntilVisible` so it reads "swipe up until visible: pre meal range label"?

src/settings/therapyScreen.js
```diff
@@ -116,7 +116,7 @@ module.exports = class TherapyScreen extends base.Screen {
         return this.deliveryLimitsScreen.InfoLabel();
     }
     async OpenDeliveryLimitsScreen() {
-        await this.SwipeUp(this.DeliveryLimitsLabel());
+        await this.SwipeUpUntil(this.DeliveryLimitsMaxBolusLabel());
```
darinkrauss: Good job on disconnecting finding labels/controls from the screen size. The new way is a lot less fragile based upon an exact number of swipes. Now, the only specific "search parameter" (other than the label/control you are looking for) is the direction (up or down). :+1:

tests/accessibility/settings/therapyScreen.tests.js
```diff
@@ -8,39 +8,42 @@ module.exports = (test) => {
...
         await expect(screen.WorkoutRangeLabel()).toBeVisible();
     });
     it('has workout range info', async () => {
         await expect(screen.WorkoutRangeInfo()).toBeVisible();
     });
-    it('swipe screen up', async () => {
-        await screen.SwipeUp(screen.BasalRateLabel(), 1);
-    });
     it('has basal rate label', async () => {
+        await screen.SwipeUpUntil(screen.BasalRateLabel());
```
darinkrauss: [nit] Aren't these two lines a bit redundant. The second line either never be hit if the label is not visible (the first line will throw if it isn't found in maxSwipes) or the second line will always succeed if the label was found by the first line. I suppose it is very clear what you are doing here by having both lines.

darinkrauss (approved):
A couple of nits, but LGTM! 🚀

tests/accessibility/settings/therapyScreen.tests.js
jh-bate: yeah it is slightly redundant but just felt better with a 'test' also included for clarity. Will probably change my mind at some stage though :)

PR #58, opened 2020-12-08
Updates to smoke_1 for Xcode 12

nhamming:
I think it makes sense to update the test piecemeal and this PR includes general updates with specific updates to make the `smoke_1` tests pass. There will be a number of PRs that merge into the `xcode-12` branch to make review easier.

Note:
- The change I needed to make to have the tests run locally is switching from `match.accessbility.TextLabel()` -> `match.Label()`. I'm still not sure why this is needed, but I want to get the tests running and then when time allows look more closely.
- A large number of the changes are from `package-lock.json` and `yarn.lock`, but this PR is still larger then one would hope. Hence why I'm breaking it into part.


ginnyyadav (commented):
Touched upon  a couple things in slack to try and kep the same config as it was before

ginnyyadav:
Great work here. thanks for updating the screens. According to detox 

```
by.label(label)
Match elements with the specified accessibility label (iOS) or content description (Android). In React Native, this corresponds to the value in the accessibilityLabel prop.

element(by.label('Welcome'));
```

so it seems you are using the right thing here.

ginnyyadav (approved):

nhamming:
> Great work here. thanks for updating the screens. According to detox
> 
> ```
> by.label(label)
> Match elements with the specified accessibility label (iOS) or content description (Android). In React Native, this corresponds to the value in the accessibilityLabel prop.
> 
> element(by.label('Welcome'));
> ```
> 
> so it seems you are using the right thing here.

I hope so. It removes the accessibility trait of `staticText`, makes the tests less specific but more flexible.

PR #62, opened 2020-12-10
updates for Xcode 12 (and smoke and error tests passing)

nhamming:
The following test specs should pass:
- smoke tests 1-3
- error tests 1-2


src/bolus/index.js
```diff
@@ -1,67 +1,199 @@
...
+  }
+  get DeliverButton() {
+    return match.accessible.Button(this.screenText.Deliver);
+  }
+  get EnterBolusButton() {
+    return match.accessible.Button(this.screenText.EnterBolus);
+  }
+  get BolusLabel() {
+    return match.accessible.TextLabel(this.screenText.Header);
+  }
```
ginnyyadav, line 29: i see that we are using match.accessible.Textlabel instead of match.label here. any eason?

src/status/header.js
```diff
@@ -1,84 +1,123 @@
...
+    ).toExist();
+    await this.CloseLoopAlert();
+  }
+  async ExpectLoopStatusInsulinAlert() {
+    await this.Loop();
+    await expect(
+      match.accessible.AlertLabel(
+        this._language.general.Alert.MissingInsulinEffects
+      )
+    ).toExist();
```
ginnyyadav, line 84: Just another question. .toExist will check if it exists, .tobBeVisible will check if it actually visible to the user on the screen. essentially it can exist but still not be visible eith .toExist. is there any reason we are using .toExist instead of .toBeVisible here?

ginnyyadav (approved):
couple questions otherwise looks good

src/bolus/index.js
nhamming: Likely this will change when I get to those tests. I haven't done a replace all, but changing as needed.

PR #63, opened 2020-12-10
Updates to functional_1 tests

nhamming:


tests/functional/index.js
```diff
@@ -57,20 +61,17 @@ module.exports = (test) => {
...
-            await test.LoopUtilities.openLoop();
+        it('can add carbs and deliver bolus', async () => {
+            await test.LoopUtilities.addCarbohydratesAndDeliverBolus(1);
         });
-        it('can add carbs', async () => {
-            await test.LoopUtilities.addCarbohydrates(5);
+    });
+    describe('bolus', () => {
+        it('can be delivered', async () => {
+            await test.LoopUtilities.deliverBolus(0.1);
```
ginnyyadav, line 74: Just making a note to ourselves here. In future we can extend this e2e test to make sure the temporary bolus banner appears while delivering and check the elements on that.

ginnyyadav (approved):
LGTM!

tests/functional/index.js
nhamming: Yeah, there are a number of UI checks that could be added for lots of different behaviours. Unfortunately, we are still building from the ground up... not adding new features. If the dev team helps with adding new features, this will make it possible for the UI tests to catch up. Otherwise it will always be lagging.

PR #67, opened 2020-12-16
[LOOP-1531] pump sim auto configures

nhamming:
https://tidepool.atlassian.net/browse/LOOP-1531

The pump simulator no longer has a setup view.


ginnyyadav (approved):
LGTM!

PR #70, opened 2021-02-12
updated to account for 3 start screens

ginnyyadav:
On the most recent release there is no longer a tidepool login screen, on the dev branch upon opening the app there is the first page of therapy settings, then on release prior there is a tidepool login screen. I went ahead and tried to account for those 3 scenarios while the onboarding is being built. This will need to change again once the onboarding is finished.


nhamming (approved):
LGTM

PR #71, opened 2021-02-25
account for no onboarding screen showing on ios13

ginnyyadav:
on ios 13 no onboading is present so i simplified this to include either swiping down the onboarding screen (if it happens to appear) or doing nothing but logging info to the console. I have tested the current implementation and it runs fine locally on ios 13.

The alternative to this would be to just remove the `skipTidepoolLogin()` entirely while we are still running on ios 13 (since onboarding will NOT show up while running on ios 13). Let me know what you think.


nhamming (approved):
LGTM

PR #72, opened 2021-03-26
iOS 14.5 (Xcode 12.5) and targeting iPhone 12 Pro

nhamming:
The large number of changes is related to `package-lock.json` and `yarn.lock`.

The real updates are:
- adjustments to the tests for iOS 14.5 (note the (hopefully temporarily) coordinate tapping to by-pass visibility checks)
- injecting allow insecure connections into the `Info.plist`
- disabling screenshots
- moving to detox `18.15.0`

I'll going to make sure LoopWorkspace scripts can handle this update before I merge.
https://github.com/tidepool-org/LoopWorkspace/pull/470


package.json
```diff
@@ -21,51 +21,31 @@
...
         "screenshot": {
-          "enabled": true,
-          "shouldTakeAutomaticSnapshots": true,
-          "keepOnlyFailedTestsArtifacts": true,
-          "takeWhen": {
-            "testStart": true,
-            "testDone": true,
-            "appNotReady": true
-          }
+          "enabled": false
```
nhamming, line 24: @Ginny91 I never use the screen shots and they are slow on circle CI. So I disabled them. However, I'm not sure this is the right choice. What do you think?

src/settings/index.js
```diff
@@ -42,13 +42,15 @@ class SettingsScreen extends base.Screen {
     async ClosedLoop() {
         const isOn = await this.IsButtonOn(this.LoopSwitchButton);
         if (!isOn) {
-            await this.LoopSwitchButton.longPress();
+            // TODO figure out why untargetted tap does not tap on the switch
```
darinkrauss, line 45: [nit] TODO?

src/settings/index.js
```diff
@@ -42,13 +42,15 @@ class SettingsScreen extends base.Screen {
...
+            await this.LoopSwitchButton.tap({"x":280,"y":20});
         }
     }
     async OpenLoop() {
         const isOn = await this.IsButtonOn(this.LoopSwitchButton);
         if (isOn) {
-            await this.LoopSwitchButton.longPress();
+            // TODO figure out why untargetted tap does not tap on the switch
```
darinkrauss, line 52: [nit] TODO?

src/settings/therapyScreen.js
```diff
@@ -164,40 +164,47 @@ module.exports = class TherapyScreen extends base.Screen {
   }
   async OpenInsulinSensitivitiesScreen() {
     await this.SwipeUpUntilVisible(this.InsulinSensitivitiesLabel);
-    await this.InsulinSensitivitiesLabel.tap();
+    // tapping coordinates by-passes visibility check
+    await this.InsulinSensitivitiesLabel.tap({"x":20,"y":20});
```
darinkrauss, line 168: [nit] How will this work if accessibility text size is changed (thus moving around all of the elements)?

darinkrauss: Or is this relative to the label so most times it would still be in the label?

src/utilities.js
```diff
@@ -96,12 +96,22 @@ module.exports = class Utilities {
...
   }
-  async dismissTidepoolLogin() {
-    try {
-      await match.Label("Welcome_1_Top").swipe('down', 'fast', 0.8);
-    } catch (e) {
-      console.log("onborading is not present on ios 13");
+  async bypassTidepoolOnboarding() {
+    // TODO figure out why hosting views make contained element not visible (and thus cannot tap)
+    // step through the onboarding screens
+    for (var pageCount = 0; pageCount < 7; pageCount++) {
```
darinkrauss, line 102: How long does this skipping take in the overall tests? If all of this is too long, we can add another way to skip everything from the first Welcome page. Perhaps not even a long press.

darinkrauss (approved):
A few comments/questions, but LGTM!

src/settings/therapyScreen.js
nhamming: It is the inner coordinates of the label. typical tap size is 40x40, so this should be good.

src/utilities.js
nhamming: in the order of seconds, where most of it is the 4 second long press.

PR #75, opened 2021-07-26
Correct failing tests

nhamming:
Note that the insulin sensitivity test are not robust and cause a timeout within detox often.

### DIFF
--- a/src/base/entryScreen.js
+++ b/src/base/entryScreen.js
@@ -29,7 +29,7 @@ class EntryScreen extends base.Screen {
   }
   get GuardrailWarningIconPicker() {
     return match.accessible
-      .Image(this.generalText.Alert.ExclamationMark)
+      .Image(this.generalText.Alert.Warning)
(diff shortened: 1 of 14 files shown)


src/base/entryScreen.js
```diff
@@ -39,7 +39,12 @@ class EntryScreen extends base.Screen {
     return match.Label(text);
   }
   async OpenPicker(value) {
-    await match.Label(String(value)).tap();
+    try {
+      await match.Label(String(value)).toBeVisible();
+      await match.Label(String(value)).tap()
+    } catch (e) {
+      await match.Label(String(value)).tap({"x":20,"y":20})
+    }
```
nhamming, line 47: @ginnyyadav  I was thinking about some generic action similar to this for by-passing visibility checks. While this might be a good idea, it would be expensive to wait for the exception each time before tapping.

ginnyyadav (approved):
LGTM!

PR #76, opened 2021-07-27
Loop restructure 1

ginnyyadav:
No functionality changes here, just initial some cleanup
- linting according to air-bnb style like Tidepool's other projects
- moving all tests into the E2E folder : this allows us to have a separation between tests and UI screens. For future development, this will make it easier to edit tests


nhamming (approved):
LGTM

I ran the smoke tests locally and they passed. I also skimmed the code changes and didn't detect any issues.

PR #77, opened 2022-01-24
LOOP-3951: Fix/update some of the UI smoke tests that had bit-rotted, enables running detox on iOS 15

rickpasetto:
https://tidepool.atlassian.net/browse/LOOP-3951

### DIFF
--- a/src/config/enUSText.js
+++ b/src/config/enUSText.js
@@ -316,11 +316,11 @@ var SettingsScreen = {
     Header: 'Delivery Limits',
     Info:
       'Maximum basal rate is the highest temporary basal rate Loop is allowed to set automatically.\n\nMaximum bolus is the highest bolus amount you can deliver at one time to cover carbs or bring down high glucose.',
-    MaxBasalRate: 'Maximum Basal Rate',
+    MaxBasalRate: 'Maximum Basal Rate, 5, U⁠/⁠hr',
     MaxBasalRateInfo:
       'Maximum Basal Rate is the highest temporary basal rate Tidepool Loop is allowed to set automatically.',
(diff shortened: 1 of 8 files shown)


nhamming (approved):
LGTM

src/config/enUSText.js
```diff
@@ -316,11 +316,11 @@ var SettingsScreen = {
     Header: 'Delivery Limits',
     Info:
       'Maximum basal rate is the highest temporary basal rate Loop is allowed to set automatically.\n\nMaximum bolus is the highest bolus amount you can deliver at one time to cover carbs or bring down high glucose.',
-    MaxBasalRate: 'Maximum Basal Rate',
+    MaxBasalRate: 'Maximum Basal Rate, 5, U⁠/⁠hr',
```
ginnyyadav, line 319: Quick question. Just wondering why this value is being hard-coded. Is it not targetable without the number (i.e.5 U/hr)? Or is it typically always going to be 5 U/hr?

ginnyyadav (approved):
LGTM. Just one question I left. Otherwise, thanks for getting these moving!!

src/config/enUSText.js
rickpasetto: Short answer: It isn't targetable unless it is the full string.  The targeting uses the accessibility element which, I imagine, for actual accessibility reasons includes the string, value, and unit.  However, this is somewhat puzzling to me because that code has been around since Nov 2020, so why is this just failing now?  That's a mystery.

ginnyyadav: Definitely strange, but thanks for explaining. Seems like detox is probably doing the "right" thing now.

PR #82, opened 2022-07-27
LOOP-4208 Upgrade detox

ps2:
https://tidepool.atlassian.net/browse/LOOP-4208

I had to dig into this to figure out why my DIY Sync changes were breaking the UITests, as I wanted to run them on my local machine, with a newer xcode version. I think I may have broken some of the reporting, but I haven't dug into that yet; just got it to the point where it was actually running locally, with a newer detox version. Wanted to save these changes for when we actually get a chance to focus on this.


ps2:
The blocking bug I encountered appears to be fixed here: https://github.com/wix/Detox/releases/tag/19.9.2

nhamming (approved):
LGTM

PR #86, opened 2023-05-23
[LOOP-4639] corrected UI test for new pump sim UI

nhamming:
https://tidepool.atlassian.net/browse/LOOP-4639


ps2 (approved):
LGTM!  Not sure if there is any way to do this, but it would be nice to be able to have UI tests not have to wait 7 seconds to to open pump setttings.

PR #90, opened 2023-09-28
Fix UI tests

ps2:


Camji55 (approved):

ginnyyadav (approved):
You definitely hit the right issue here. Should fix it.

PR #91, opened 2023-09-29
LOOP-4735 Make UI Tests less flakey

ps2:
Add an additional swipe up, even if the element is visible, to account for home bar


ginnyyadav (approved):
LGTM 🤞 🚢

